PR #20, opened 2022-07-11
feat: add Channel widget

fdrgsp:
This PR adds the Channel widget and the tests.


src/pymmcore_widgets/_util.py
```diff
@@ -0,0 +1,41 @@
...
+        btn_box.rejected.connect(self.reject)
+
+        self.setLayout(QVBoxLayout())
+        if text:
+            self.layout().addWidget(QLabel(text))
+        self.layout().addWidget(self._combo)
+        self.layout().addWidget(btn_box)
+
+    def _currentText(self) -> str:
+        return self._combo.currentText()  # type: ignore [no-any-return]
```
ianhi, line 41: This should probably be public? Or else the `self.accept` and `self.reject` methods should give the current value as part of their signal

PR #21, opened 2022-07-11
feat: add Objectives widget

fdrgsp:
This PR adds the Objectives widget and the tests.

### DIFF
--- a/src/pymmcore_widgets/_objective_widget.py
+++ b/src/pymmcore_widgets/_objective_widget.py
@@ -96,8 +96,7 @@ class ObjectivesWidget(QWidget):
 
 
 class _ObjectiveStateWidget(StateDeviceWidget):
-    """
-    Subclass of StateDeviceWidget.
+    """Subclass of StateDeviceWidget.
 
     Drops/raises stage when changing objective.
     """
--- a/src/pymmcore_widgets/_objective_widget.py
+++ b/src/pymmcore_widgets/_objective_widget.py
@@ -1,0 +1,123 @@
+from typing import Optional, Union
+
+from pymmcore_plus import CMMCorePlus
+from qtpy.QtWidgets import QComboBox, QHBoxLayout, QLabel, QSizePolicy, QWidget
+
+from ._core import get_core_singleton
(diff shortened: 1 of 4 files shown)


src/pymmcore_widgets/_objective_widget.py
```diff
@@ -0,0 +1,123 @@
...
+        self._combo.setParent(None)
+        self._combo.deleteLater()
+
+    def _create_objective_combo(
+        self, device_label: Union[str, None]
+    ) -> Union[StateDeviceWidget, QComboBox]:
+        if device_label:
+            combo = _ObjectiveStateWidget(device_label, mmcore=self._mmc)
+            combo.setMinimumWidth(285)
+            combo._combo.currentIndexChanged.connect(self._on_obj_changed)
```
tlambert03, line 88: this line is new... was there an issue in the old one that this fixes?

src/pymmcore_widgets/_objective_widget.py
```diff
@@ -0,0 +1,123 @@
...
+            combo.setEnabled(False)
+        return combo
+
+    def _on_obj_changed(self) -> None:
+        self._mmc.events.pixelSizeChanged.emit(self._mmc.getPixelSizeUm() or 0.0)
+
+
+class _ObjectiveStateWidget(StateDeviceWidget):
+    """
+    Subclass of StateDeviceWidget.
```
tlambert03, line 100: put summary comment on the first line:

```suggestion
    """Subclass of StateDeviceWidget.
```

Also ... try to avoid comments that literally say what the code says (the line above says that this is just a subclass of state device widget).  Either skip the comment entirely for a private class, or give a one-line explanation for what it _does_ (not what it is)

tlambert03 (approved):

src/pymmcore_widgets/_objective_widget.py
fdrgsp: I added this line because `pymmcore` does not emit the `pixelSizeChanged` signal when the objective is changed (it sets the new pixel size without emitting the signal).

PR #22, opened 2022-07-11
feat: add Slider Dialog widget

fdrgsp:
This PR adds the Slider Dialog widget and the tests.

### DIFF
--- a/src/pymmcore_widgets/_slider_dialog_widget.py
+++ b/src/pymmcore_widgets/_slider_dialog_widget.py
@@ -16,6 +16,11 @@ class SliderDialog(QDialog):
     ----------
     property_regex : str
         a regex pattern to use to select property names to show.
+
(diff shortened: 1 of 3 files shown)


src/pymmcore_widgets/_slider_dialog_widget.py
```diff
@@ -0,0 +1,54 @@
...
+from ._property_widget import PropertyWidget
+
+
+class SliderDialog(QDialog):
+    """A Dialog that shows range-based properties (such as light sources) as sliders.
+
+    Parameters
+    ----------
+    property_regex : str
+        a regex pattern to use to select property names to show.
```
tlambert03, line 18: hard to tell without looking at the code what this parameter is supposed to be.  A simple example or description of what the regex needs to contain would be good

tlambert03 (approved):

PR #23, opened 2022-07-11
feat: add Load System cfg widget

fdrgsp:
This PR adds the widget to load the micromanager system cfg and the tests.

### DIFF
--- a/src/pymmcore_widgets/_load_system_cfg_widget.py
+++ b/src/pymmcore_widgets/_load_system_cfg_widget.py
@@ -1,15 +1,23 @@
 from typing import Optional
 
+from pymmcore_plus import CMMCorePlus
 from qtpy import QtWidgets as QtW
 
-from ._core import load_system_config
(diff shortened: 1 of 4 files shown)


src/pymmcore_widgets/_load_system_cfg_widget.py
```diff
@@ -0,0 +1,40 @@
...
+
+from qtpy import QtWidgets as QtW
+
+from ._core import load_system_config
+
+
+class ConfigurationWidget(QtW.QGroupBox):
+    """Widget to select and load MM configuration."""
+
+    def __init__(self, parent: Optional[QtW.QWidget] = None) -> None:
```
tlambert03, line 11: we should be allowing all of these widgets to take in a (non-global) core object.  so `load_system_configuration` should be modified to take an optional `mmcore` argument, and this widget should also take one in the init

fdrgsp: @tlambert03 fixed

PR #24, opened 2022-07-11
feat: add Shutter widget

fdrgsp:
This PR adds a general Shutter widget and the tests.

---

Screen Shot 2022-10-07 at 3 51 04 PM

Screen Shot 2022-10-07 at 3 51 26 PM

Screen Shot 2022-10-07 at 3 51 37 PM

### DIFF
--- a/src/pymmcore_widgets/_shutter_widget.py
+++ b/src/pymmcore_widgets/_shutter_widget.py
@@ -316,14 +316,3 @@ class ShuttersWidget(QtW.QWidget):
         self._mmc.events.startSequenceAcquisition.disconnect(self._on_seq_started)
         self._mmc.events.stopSequenceAcquisition.disconnect(self._on_seq_stopped)
         self._mmc.events.imageSnapped.disconnect(self._on_seq_stopped)
-
-
-# from pymmcore_widgets._core import get_core_singleton
-# from pymmcore_widgets._shutter_widget import ShuttersWidget
-# from fonticon_mdi6 import MDI6
-# from qtpy.QtGui import QIcon
-# i = QIcon("/Users/FG/Desktop/down.png")
-# mmc = get_core_singleton()
-# mmc.loadSystemConfiguration()
-# s = ShuttersWidget('Shutter')
-# s.show()
--- a/src/pymmcore_widgets/_shutter_widget.py
+++ b/src/pymmcore_widgets/_shutter_widget.py
@@ -10,8 +10,6 @@ from qtpy.QtGui import QColor
 from superqt.fonticon import icon
 from superqt.utils import signals_blocked
 
-from ._core import get_core_singleton
-
 COLOR_TYPE = Union[
     QColor,
     int,
@@ -75,7 +73,7 @@ class ShuttersWidget(QtW.QWidget):
     ) -> None:
         super().__init__(parent)
 
-        self._mmc = mmcore or get_core_singleton()
+        self._mmc = mmcore or CMMCorePlus.instance()
 
         self.shutter_device = shutter_device
         self._is_multiShutter = False
--- a/src/pymmcore_widgets/_shutter_widget.py
+++ b/src/pymmcore_widgets/_shutter_widget.py
@@ -73,6 +73,7 @@ class ShuttersWidget(QtW.QWidget):
         self._mmc.events.startSequenceAcquisition.connect(self._on_seq_started)
         self._mmc.events.stopSequenceAcquisition.connect(self._on_seq_stopped)
         self._mmc.events.imageSnapped.connect(self._on_seq_stopped)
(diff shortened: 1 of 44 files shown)


src/pymmcore_widgets/_shutter_widget.py
```diff
@@ -0,0 +1,232 @@
...
+        shutter_device: str,
+        autoshutter: bool = True,
+        button_text_open_closed: Tuple[str, str] = (None, None),  # type: ignore
+        icon_open_closed: Tuple[QIcon, QIcon] = (
+            MDI6.hexagon_outline,
+            MDI6.hexagon_slice_6,
+        ),
+        icon_size: int = 25,
+        icon_color_open_closed: tuple[COLOR_TYPE, COLOR_TYPE] = ("", ""),
+        text_color_combo: COLOR_TYPE = "",
```
tlambert03, line 61: i dislike having all of these style-related things in the `__init__` method.  Also, it looks like at least `text_color_combo` is never used anywhere.  Can you

1) remove these from the init, and expose setter methods for the properties that you feel _must_ be exposed (but don't recreate any setters than Qt already provides... we don't want to create any additional APIs that aren't strictly necessary)
2) review them all to make sure they are used?

tlambert03: side note: this won't be necessary after you remove all the params from the `__init__`, but if you have this many parameters in a function call, you should generally try to make them all keyword only (i.e. move the `*` farther up in the call).  This allows you to reposition them later without breaking anyone's usage (if they happened to be calling them positionally)

src/pymmcore_widgets/_shutter_widget.py
```diff
@@ -0,0 +1,327 @@
...
+        sizepolicy_btn = QtW.QSizePolicy(QtW.QSizePolicy.Fixed, QtW.QSizePolicy.Fixed)
+        self.shutter_button.setSizePolicy(sizepolicy_btn)
+        self.shutter_button.setIcon(
+            icon(self._icon_closed, color=self._icon_color_closed)
+        )
+        self.shutter_button.setIconSize(QSize(self._icon_size, self._icon_size))
+        self.shutter_button.clicked.connect(self._on_shutter_btn_clicked)
+        main_layout.addWidget(self.shutter_button)
+
+        if self.autoshutter:
```
tlambert03, line 200: rather than actually changing the attributes this widget has, lets just do `if not self.autoshutter: self.autoshutter_checkbox.hide()`

fdrgsp:
@tlambert03  I wanted to remove the properties (and setter methods) from the `shutter` widget and to leave them as attributes. I also want to create a public method called `update_widget_style()` that you can call after change any of the attributes if you want to immediately apply the style changed:

```
def update_widget_style(self) -> None:
        """
        Update the style of the shutter widget.
        
        Call after changing any of the shutter attributes to apply the new style.
        """

        if int(self._mmc.getProperty(self.shutter_device, "State")) == 1:
            self.shutter_button.setIcon(icon(self.icon_open, color=self.icon_color_open))
            self.shutter_button.setText(self.button_text_open)
        else:
            self.shutter_button.setIcon(icon(self.icon_closed, color=self.icon_color_closed))
            self.shutter_button.setText(self.button_text_closed)

        self.shutter_button.setIconSize(QSize(self.icon_size, self.icon_size))
```

What do you think about this solution?

tlambert03:
> `update_widget_style`

that looks like a method that should be public, but rather should be called behind the scenes when any of the data required to update has changed.  In other words, in what scenario would an end user _use_ this method?  would it be something they should always call right after setting one of the other attributes?  If so, shouldn't we just call it for them?

tests/test_shutter_widget.py
```diff
@@ -0,0 +1,150 @@
...
+import pytest
+
+from pymmcore_widgets._shutter_widget import ShuttersWidget
+
+if TYPE_CHECKING:
+    from pymmcore_plus import CMMCorePlus
+    from pytestqt.qtbot import QtBot
+
+
+def test_shutter_widget(qtbot: QtBot, global_mmcore: CMMCorePlus):
```
tlambert03, line 14: this test is too big.  test smaller things... ie. test:
1.  instantiation on it's own
2. shutter vs multishutter in different tests
3. autoshutter true vs false in different tests
4. each type of global change

... basically, anywhere you'd be tempted to write "# test xyz" ... try to make a new test.  Imagine the test failing, and then trying to figure out _why_ it failed... (smaller tests are easier to get right to the failure))

tlambert03: i think this is the only thing this PR is waiting on

fdrgsp: @tlambert03 should be ok now.

tlambert03 (approved):
LGTM!

PR #25, opened 2022-07-11
feat: add pixel size widget

fdrgsp:
This PR adds a widget to set the pixel size and the tests.

*it should to be merged after [#21](https://github.com/pymmcore-plus/pymmcore-plus/pull/143)

Screen Shot 2022-09-22 at 4 44 21 PM


tlambert03:
taking a quick look through this...  (partially to determine the need for the various signals added in https://github.com/pymmcore-plus/pymmcore-plus/pull/143/)

my first thought is that it has grown to be rather complicated.  I'm not certain yet, but it kinda looks like you're trying to be very granular with the possible things that could happen.  But I feel like you could save yourself a whole lot of complexity here by having a single `rebuild` method that just clears the table and creates it fresh from the state of core.  (and then you could have a single signal emitted from core indicating that something/anything about the pixel config has changed).

if you're thinking that sounds "wasteful" ... it probably only takes a couple hundred ms... and it will be relatively rare that a programatic change will occur *as* someone is using it in the gui...

src/pymmcore_widgets/_pixel_size_widget.py
```diff
@@ -0,0 +1,590 @@
...
+        *,
+        mmcore: Optional[CMMCorePlus] = None,
+    ) -> None:
+        super().__init__(parent)
+
+        self.setWindowTitle("Set Image Pixel Size")
+
+        self._mmc = mmcore or CMMCorePlus.instance()
+
+        self._magnification: List[List[str, float]] = []  # type: ignore
```
tlambert03, line 74: sometimes the type linter is telling you something you shouldn't ignore :joy:  `List[]` just doesn't take two arguments... if you want to specify the items in a list, you have to use `Tuple[str, float]`...

however... this _really_ kinda looks like it should just be a dictionary, (mapping objective label to magnification), no?

tlambert03: (actually... it looks like this is unused altogether?? :joy:)

tlambert03:
I started to refactor this a bit... but it got hairy.  Where I was going with it was to break it up into smaller concepts.  Like:

1. a function that will retrieve all of the data from the core needed to build a table (but doesn't actually build the table)

<details>

```python
def _get_pixel_info(
    core: CMMCorePlus = None, obj_dev: str = "Objective"
) -> List[Dict[str, Any]]:
    """Returns a list of records, that can be used to build a table:
    
    [
        {'objective': 'Something', 'resolutionID": 'something', ...},
         ...
    ]
    """
    core = core or CMMCorePlus.instance()

    # e.g.
    # { 'Nikon 20X Plan Fluor ELWD': ('Res20x', 0.5) }
    obj_cfg_px: Dict[str, Tuple[str, float]] = {}
    for cfg in core.getAvailablePixelSizeConfigs():
        obj = next(iter(core.getPixelSizeConfigData(cfg)))[2]
        obj_cfg_px[obj] = (cfg, core.getPixelSizeUmByID(cfg))

    # get pixel info for each objective
    rows = []
    mag_factor = core.getMagnificationFactor()
    cam_px_guess = 0.0
    for obj_label in core.getStateLabels(obj_dev):
        res_id, px_value = obj_cfg_px.get(obj_label, ("", 0.0))
        tot_mag = _guess_mag(obj_label) * mag_factor
        rows.append(
            {
                "objective": obj_label,
                "resolutionID": res_id,
                "cameraPixelSize": 0,
                "magnification": tot_mag,
                "imagePixelSize": px_value,
            }
        )
        if not cam_px_guess and (tot_mag and px_value):
            cam_px_guess = tot_mag * px_value

    if cam_px_guess:
        for row in rows:
            row["cameraPixelSize"] = cam_px_guess

    return rows
```

</details>

2. a method on the table that can take that info and rebuild the table:

<details>

```python

class PixelSizeTable(QtW.QTableWidget):
    """Create a Table to set pixel size configurations."""

    HEADERS = {
        "objective": "Objective",
        "resolutionID": "Resolution ID",
        "cameraPixelSize": "Camera Pixel Size (um)",
        "magnification": "Magnification",
        "imagePixelSize": "Image Pixel Size (um)",
        "delete": "Delete",
    }

    def __init__(self, core: CMMCorePlus, parent: Optional[QtW.QWidget] = None) -> None:
        super().__init__(parent)
        self._mmc = core
        hh = self.horizontalHeader()
        hh.setSectionResizeMode(hh.Stretch)
        hh.setDefaultAlignment(Qt.AlignHCenter)
        vh = self.verticalHeader()
        vh.setVisible(False)
        vh.setSectionResizeMode(hh.Stretch)
        self.setSelectionBehavior(QtW.QAbstractItemView.SelectItems)
        self.setDragDropMode(QtW.QAbstractItemView.NoDragDrop)

    def _rebuild(self, obj_dev: str = "Objective") -> None:
        records = _get_pixel_info(self._mmc, obj_dev)

        self.clear()
        self.setRowCount(len(records))
        self.setColumnCount(len(self.HEADERS))  # +1 for delete button
        self.setHorizontalHeaderLabels(list(self.HEADERS.values()))

        for row, record in enumerate(records):
            self._populate_row(row, record)

    def _populate_row(self, row: int, record: dict) -> None:
        """populate a row with widgets."""
        for col, key in enumerate(self.HEADERS):
            if key == "delete":
                wdg = self._new_delete_btn()
            else:
                wdg = self._new_line_edit(record[key])
                wdg.editingFinished.connect(self._on_text_entered)
            self.setCellWidget(row, col, wdg)

    def _new_line_edit(self, value: Any, read_only: bool = False) -> QtW.QLineEdit:
        item = QtW.QLineEdit(str(value))
        item.setAlignment(Qt.AlignCenter)
        item.setFrame(False)
        if read_only:
            item.setReadOnly(True)
            item.setStyleSheet("color:magenta")
        else:
            item.setReadOnly(False)
            item.setStyleSheet("")

        if isinstance(value, (int, float)):
            item.setValidator(QDoubleValidator())
        return item

    def _new_delete_btn(self) -> QtW.QWidget:
        wdg = QtW.QWidget()
        layout = QtW.QHBoxLayout()
        layout.setSpacing(0)
        layout.setContentsMargins(0, 0, 0, 0)
        layout.setAlignment(Qt.AlignCenter)
        wdg.setLayout(layout)

        btn = QtW.QPushButton(text="Delete")
        btn.setFixedWidth(70)
        btn.setSizePolicy(QtW.QSizePolicy.Fixed, QtW.QSizePolicy.Fixed)
        btn.setToolTip("Delete configuration.")
        btn.clicked.connect(self._delete_cfg)

        btn.setAutoDefault(False)
        layout.addWidget(btn)
        return wdg

    def _delete_cfg(self) -> None:
        w = self.sender().parent()
        row = self.indexAt(w.pos()).row()

        resId_edit = cast("QtW.QLineEdit", self.cellWidget(row, RESOLUTION_ID))
        if resId_edit.text() in self._mmc.getAvailablePixelSizeConfigs():
            with signals_blocked(resId_edit):
                resId_edit.setText("None")
            im_px_item = self.cellWidget(row, IMAGE_PX_SIZE)
            with signals_blocked(im_px_item):
                im_px_item.setText("0.0")
            return
        self._mmc.deletePixelSizeConfig(resId_edit.text())

    def _on_text_entered(self):
        sender = self.sender()
        row = self.indexAt(sender.parent().pos()).row()
        col = self.indexAt(sender.parent().pos()).col()
        ...
```

</details>

(note that you can also use `indexAt(sender().parent().pos())` rather than using QProperties)

3. then ... in your main widget... you can just make any "external" event rebuild the table:

<details>

```python
class PixelSizeWidget(QtW.QDialog):
        self._mmc.events.systemConfigurationLoaded.connect(self._rebuild)
        # self._mmc.events.pixelSizeDefined.connect(self._rebuild)
        # self._mmc.events.pixelSizeSet.connect(self._rebuild)
        # self._mmc.events.pixelSizeDeleted.connect(self._rebuild)
        self._rebuild()

    def _rebuild(self) -> None:
        if not (obj_dev := self.objective_device):
            return
        self.table._rebuild(obj_dev)

        h = self.sizeHint().height()
        self.resize(self.sizeHint().width() * 2, int(h + (h * 20 / 100)))
```

</details>

tlambert03:
side note, rather than instantiating a `ObjectivesWidget()`  just to use it's guess objective method, you should make a helper function that both widgets use:

```python
# in _util.py

def guess_objective_or_prompt(core=None, parent=None) -> Optional[str]:
    """Try to update the list of objective choices.

    1. get a list of potential objective devices from pymmcore
    2. if there is only one, use it, if there are >1, show a dialog box
    """
    core = core or CMMCorePlus.instance()
    candidates = core.guessObjectiveDevices()
    if len(candidates) == 1:
        return candidates[0]
    elif candidates:
        dialog = ComboMessageBox(candidates, "Select Objective Device:", parent=parent)
        if dialog.exec_() == dialog.DialogCode.Accepted:
            return dialog.currentText()
    return None

```

src/pymmcore_widgets/_pixel_size_widget.py
```diff
@@ -0,0 +1,590 @@
...
+        item.setFrame(False)
+        if self._is_read_only(col):
+            item.setReadOnly(True)
+            item.setStyleSheet("color:magenta")
+        else:
+            item.setReadOnly(False)
+            item.setStyleSheet("")
+
+        if col in {2, 3, 4}:
+            regex = QRegularExpression(FLOAT_REGEX)
```
tlambert03, line 257: there is a QDoubleValidator made just for this: 
https://doc.qt.io/qt-6/qdoublevalidator.html

tests/conftest.py
```diff
@@ -18,3 +21,13 @@ def global_mmcore(core):
     assert mmc == core
     mmc.loadSystemConfiguration(str(Path(__file__).parent / "test_config.cfg"))
     return mmc
+
+
+@pytest.fixture()
+def px_wdg(global_mmcore, qtbot: QtBot):
```
tlambert03, line 27: unless you need to use this fixture in multiple different test files, it should be placed in the test file where it's being used.  (i.e.  move this to `test_table_pixel_size_widget`)

tests/test_table_pixel_size_widget.py
```diff
@@ -0,0 +1,225 @@
...
+ROW = 0  # "Nikon 40X Plan Fluor ELWD"
+
+
+def _get_wdg(table):
+    obj = table.item(ROW, OBJECTIVE_LABEL).text()
+    resID = cast(QLineEdit, table.cellWidget(ROW, RESOLUTION_ID))
+    mag = cast(QLineEdit, table.cellWidget(ROW, MAGNIFICATION))
+    cam_px = cast(QLineEdit, table.cellWidget(ROW, CAMERA_PX_SIZE))
+    img_px = cast(QLineEdit, table.cellWidget(ROW, IMAGE_PX_SIZE))
+    return obj, resID, mag, cam_px, img_px
```
tlambert03, line 25: i'm curious... why did you do these casts?  because you're simply returning them, and because `_get_wdg` itself has no return annotations, they would appear to be doing nothing?

tests/test_table_pixel_size_widget.py
```diff
@@ -0,0 +1,225 @@
...
+
+
+def test_change_magnification(qtbot: QtBot, px_wdg):
+    _, table, _, mmc = px_wdg
+
+    _, _, mag, cam_px, img_px = _get_wdg(table)
+
+    mag.setText("50.0")
+    with qtbot.waitSignal(mag.editingFinished):
+        mag.editingFinished.emit()
```
tlambert03, line 63: these patterns (where you await the very signal you are explicitly emitting) are a bit strange.  did you come to this out of necessity?

Normally, `waitSignal` is used when you expect some signal to mark the end of some other action:

```python
with qtbot.waitSignal(widget.valueChanged):
    widget.push_unrelated_button()
```

so, why isn't it sufficient just to use `mag.setText("50.0")`?   if someone programatically did that and then didn't follow it up with `editingFinished.emit()` would things be out of sync?

fdrgsp: @tlambert03  I've use this pattern because `wdg.setText()` does not emit the `editingFinished` signal. I preferred to use the use the `editingFinished` signal for the QLineEdit table widgets instead of the `textChanged` signal so that the signal is emitted only when the user press enter and not every time is typing. 
Do you think we should I switch to the `textChanged` signal?

tlambert03: nah, the goal of not over-emitting on every letter is definitely a good one.  I suppose we can just leave the potential "buggish" thing of someone manipulating the GUI programatically with `setText`... or at least cross that bridge later

tlambert03 (approved):
hooray!  great job!

PR #26, opened 2022-07-11
feat: add MultiD widget

fdrgsp:
This PR adds a MultiD Widget and the tests.

Compared to the previous version we had on `napari-micromanager`, here I have added:
- position names in the stage position table.
- a small widget that generates a grid of coordinates (open using the Grid button in the stage position table) that are then added to the stage position table.

---
Screen Shot 2022-10-02 at 3 56 35 PM


ianhi (changes requested):

src/pymmcore_widgets/_mda_widget/_mda_gui.py
```diff
@@ -0,0 +1,499 @@
...
+        acq_wdg_layout = QHBoxLayout()
+        acq_wdg_layout.setSpacing(0)
+        acq_wdg_layout.setContentsMargins(0, 0, 0, 0)
+        acq_wdg.setLayout(acq_wdg_layout)
+        acquisition_order_label = QLabel(text="Acquisition Order:")
+        lbl_sizepolicy = QSizePolicy(QSizePolicy.Fixed, QSizePolicy.Fixed)
+        acquisition_order_label.setSizePolicy(lbl_sizepolicy)
+        self.acquisition_order_comboBox = QComboBox()
+        self.acquisition_order_comboBox.setMinimumWidth(100)
+        self.acquisition_order_comboBox.addItems(["tpzc", "tpcz", "ptzc", "ptcz"])
```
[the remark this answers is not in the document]
tlambert03: btw @fdrgsp ... do you know that you can click the `commit suggestion` button and have these applied automatically?  it does make it a bit easier to "confirm" that the change as actually been incorporated

fdrgsp:
@tlambert03 at the moment the `run_Button` is connected to  `_on_run_clicked` and I think this can be a problem if someone wants to customize the function of this button. And I agree that is is not ideal to have the `run_Button` not connected to anything. But I also think is it good to already have  the run, cancel and pause buttons already in for an easier use of the widget.

What about if we remove the `run_Button` and the `_on_run_clicked` method, we make public the `_get_state` method and in the doctoring we say something like this?

```
class MultiDWidget(MultiDWidgetGui):
    """Multi-dimensional acquisition Widget.

    Parameters
    ----------
    parent : Optional[QWidget]
        Optional parent widget, by default None
    mmcore: Optional[CMMCorePlus]
        Optional `CMMCorePlus` micromanager core.
        By default, None. If not specified, the widget will use the active
        (or create a new) `CMMCorePlus.instance()`.
    
    The `MultiDWidget` itself does not "run" an experiment, it simply provides a GUI
    to construct a `useq.MDASequence` object through its `get_state` method.

    The generated `useq.MDASequence` can then be passed to the `CMMCorePlus.run_mda`
    method to actually execute the acquisition.

    A "build and run" method can be connected to the `run_Button` of this widget
    to "build and run" the acquisition when it is clicked.

    Example:

    # create Qt Application
    app = QApplication([])

    # create CMMCore instance and load the default config
    mmc = CMMCorePlus.instance()
    mmc.loadSystemConfiguration()

    # create MultiDWidget
    mda = MultiDWidget()

    # define the "build and run" method
    def _on_run_clicked() -> None:
        # construct a `useq.MDASequence` object from the values inserted in the widget
        experiment = mda.get_state()
        # run the MDA experiment asynchronously
        mmc.run_mda(experiment)
        return
    
    # connect the build and run" method to the run_Button
    mda.run_Button.clicked.connect(_on_run_clicked)

    # execute Qt Application
    app.exec_()
    """
```

tlambert03:
how about adding an __init__ parameter `include_run_button` that defaults to False?

src/pymmcore_widgets/_mda_widget/_mda_widget.py
```diff
@@ -537,7 +537,7 @@ def set_state(self, state: dict | MDASequence | str | Path) -> None:
         else:
             self.stage_pos_groupBox.setChecked(False)
 
-    def _get_state(self) -> MDASequence:
+    def get_state(self) -> MDASequence:
```
tlambert03, line 540: curious why you made this private.  this would appear to be the one main "public" kinda thing from this widget.  It's essentially the main way for a user to get the MDASequence out right?  Are you replacing it with something else?

tlambert03: me=dumb... this is *making* it public

tlambert03 (approved):
LGTM!

PR #30, opened 2022-07-13
feat/test: move SNAP button style __init__ args to properties + fix test

fdrgsp:
This PR removes style `args` in the `__init__` method  for the `snap button` and sets them as properties + tests update.

### DIFF
--- a/src/pymmcore_widgets/_snap_button_widget.py
+++ b/src/pymmcore_widgets/_snap_button_widget.py
@@ -4,7 +4,7 @@ from fonticon_mdi6 import MDI6
 from pymmcore_plus import CMMCorePlus
 from qtpy.QtCore import QSize, Qt
 from qtpy.QtGui import QColor
-from qtpy.QtWidgets import QPushButton, QSizePolicy
+from qtpy.QtWidgets import QPushButton, QSizePolicy, QWidget
 from superqt.fonticon import icon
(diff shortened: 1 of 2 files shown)


src/pymmcore_widgets/_snap_button_widget.py
```diff
@@ -65,11 +66,40 @@ def __init__(
...
+
+    @property
+    def icon_color(self) -> COLOR_TYPES:
+        """Set the snap button icon color."""
+        return self._icon_color
+
+    @icon_color.setter
+    def icon_color(self, color: COLOR_TYPES) -> None:
+        self.setIcon(icon(MDI6.camera_outline, color=color))
+        self._icon_color = color
```
tlambert03, line 97: So these are all cases where I think the new public methods are unnecessary.  This widget already _is_ a QPushButton, and all you're doing with `button_text`, `icon_size`, etc..., is just turning around and calling Qt's `self.setSameDarnThing`.  In the other example, you had a compound `QWidget` that didn't have things like `setText`, etc... but here you do.  So, this is an example of where I wouldn't create any additional public API where one already exists.

tlambert03:
> feat/test

can't have both in a semantic commit

fdrgsp:
@tlambert03  I close this and open #39 with the proper fix.

PR #31, opened 2022-07-13
fix: move LIVE button __init__ args to properties + update test

fdrgsp:
This PR removes style `args` in the `__init__` method for the `live button` and sets them as `properties` + tests update.

### DIFF
--- a/src/pymmcore_widgets/_live_button_widget.py
+++ b/src/pymmcore_widgets/_live_button_widget.py
@@ -50,7 +50,7 @@ class LiveButton(QPushButton):
         super().__init__(parent)
 
         self._mmc = mmcore or CMMCorePlus.instance()
-        
+
         self._camera = self._mmc.getCameraDevice()
 
         self._button_text_on: str = "Live"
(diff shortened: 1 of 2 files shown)


tlambert03 (commented):
same comment as the other PR, I do think you should remove these from the `__init__`, but i don't think you should create new public getters/setters that simply turn around and call `self.doSameThing`

src/pymmcore_widgets/_live_button_widget.py
```diff
@@ -71,11 +78,65 @@ def __init__(
...
             self.setEnabled(True)
 
+    @property
+    def button_text_on(self) -> str:
+        """Set the live button text for when live mode is on."""
+        return self._button_text_on
+
+    @button_text_on.setter
+    def button_text_on(self, text: str) -> None:
+        if not self._mmc.isSequenceRunning(self._camera):
```
tlambert03, line 88: need to think about this one...

tlambert03 (approved):

PR #38, opened 2022-07-14
fix: fix pre-tests

ianhi:
See: https://github.com/ianhi/pymmcore-widgets/actions/runs/2671576282

Closes: #35 
Closes: #36 
Closes: #37


tlambert03:
actually... wait, why did you remove linux?

tlambert03:
oh ... cause it was never there before anyway, huh?

ianhi:
Yeah we don't have it on even pymmcore-plus because it'll fail on loading the config bc we don't install micromanager on linux

PR #47, opened 2022-08-02
feat: Add Camera ROI widget + test

fdrgsp:
This PR adds a new widget to dynamically control the camera ROI (`pymmcore setRoi()`).

This widget interacts with `pymmcore-plus` through the pymmcore-plus signal: `roiSet`.

Every time  any of the ROI settings is changed, a signal (`roiInfo`) is emitted containing the ROI info. It can be used to pair this widget with a `napari` layer in `napari-micromanager` (e.g. https://github.com/pymmcore-plus/napari-micromanager/pull/200).

---

Screen Shot 2022-10-01 at 4 43 56 PM

Screen Shot 2022-10-01 at 4 44 10 PM

Screen Shot 2022-10-01 at 4 44 23 PM


tlambert03:
note, the signal name was changed to `roiSet` ... so this will need updating

tlambert03:
looking good. my only confusion is about the intention of the "crop" button and how it relates to the combobox.

from the code, I see that clicking the button is the only way to actually communicate the change to micromanager, but I think the relationship with the combobox might be a bit confusing:

1. someone selects, say 256 x 256
2. they see the "Size: 256 px * 256 px change..."... but  presumably nothing has happened yet.
3. So they click the "crop" button...
4. but now the dropdown no longer says 256 x 256 ... which is slightly confusing.

I also think the text "ROI" is a bit confusing.   _all_ of the dropdown options are ROIs no?  perhaps "Custom ROI"?  
also, I would change all of the string instances of `"ROI"` to a global variable `CUSTOM = "Custom"` or something like that.... same with `"Full"`

I think the name of the signal `roiInfo` needs to change to reflect _when_ the signal is emitted (not _what_ it contains).  So, perhaps `roiChanged`.

fdrgsp:
@tlambert03 it should be ok now.

src/pymmcore_widgets/_camera_roi_widget.py
```diff
@@ -0,0 +1,474 @@
...
+    ----------
+    parent : Optional[QWidget]
+        Optional parent widget, by default None
+    mmcore: Optional[CMMCorePlus]
+        Optional `CMMCorePlus` micromanager core.
+        By default, None. If not specified, the widget will use the active
+        (or create a new) `CMMCorePlus.instance()`.
+    """
+
+    roiChanged = Signal(int, int, int, int, str)
```
tlambert03, line 47:
```suggestion
    # (x, y, width, height, comboBoxText)
    roiChanged = Signal(int, int, int, int, str)
```

src/pymmcore_widgets/_camera_roi_widget.py
```diff
@@ -0,0 +1,475 @@
...
+fixed_sizepolicy = QSizePolicy(QSizePolicy.Fixed, QSizePolicy.Fixed)
+FULL = "Full Chip"
+CUSTOM_ROI = "Custom ROI"
+
+
+class CameraRoiWidget(QWidget):
+    """
+    A Widget to control the camera device ROI.
+
+    When the ROI changes, the roiChanged Signal is emitted.
```
tlambert03, line 35:
```suggestion
    When the ROI changes, the roiChanged Signal is emitted with the current
    (x, y, width, height, comboBoxText)
```

tlambert03:
does this PR require `pymmcore-plus>=0.5.0` (for the roiSet signal?)  if so, that should be added to pyproject.toml

tlambert03:
hooray!

PR #48, opened 2022-09-05
feat: update GroupPresetTableWidget + PresetsWidget + tests

fdrgsp:
This PR add the ability to create, delete or modify `groups` and `presets` directly from the `GroupPresetTableWidget`.

Uses the new sets of `pymmcore-plus` signals and methods from https://github.com/pymmcore-plus/pymmcore-plus/pull/139. 

*** to do: update pymmcore-plus pyproject.toml when https://github.com/pymmcore-plus/pymmcore-plus/pull/139 is merged.

Screen Shot 2022-09-05 at 7 04 10 PM

Screen Shot 2022-09-05 at 7 04 36 PM

Screen Shot 2022-09-05 at 7 04 51 PM

Screen Shot 2022-09-05 at 7 05 05 PM

Screen Shot 2022-09-05 at 7 05 17 PM

### DIFF
--- a/src/pymmcore_widgets/_group_preset_widget/_add_first_preset_widget.py
+++ b/src/pymmcore_widgets/_group_preset_widget/_add_first_preset_widget.py
@@ -1,4 +1,3 @@
-import warnings
 from typing import Optional
 
 from qtpy.QtCore import Qt
@@ -17,18 +16,30 @@ from qtpy.QtWidgets import (
     QWidget,
 )
 
+# from ._core import get_core_singleton
+# from ._property_widget import PropertyWidget
+# from ._add_first_preset_widget import AddFirstPresetWidget
 from pymmcore_widgets._core import get_core_singleton
 from pymmcore_widgets._property_widget import PropertyWidget
 
 
-class AddPresetWidget(QDialog):
-    """A widget to add presets to a specified group."""
+class AddFirstPresetWidget(QDialog):
+    """A widget to create the first specified group's preset."""
 
-    def __init__(self, group: str, *, parent: Optional[QWidget] = None) -> None:
+    def __init__(
+        self,
+        group: str,
+        preset: str,
+        dev_prop_val_list: list,
+        *,
+        parent: Optional[QWidget] = None,
+    ) -> None:
         super().__init__(parent)
 
         self._mmc = get_core_singleton()
         self._group = group
+        self._preset = preset
+        self._dev_prop_val_list = dev_prop_val_list
 
         self._create_gui()
 
@@ -75,6 +86,7 @@ class AddPresetWidget(QDialog):
         ps_lbl = QLabel(text="Preset:")
         ps_lbl.setSizePolicy(lbl_sizepolicy)
         self.preset_name_lineedit = QLineEdit()
+        self.preset_name_lineedit.setText(f"{self._preset}")
 
         spacer = QSpacerItem(30, 10, QSizePolicy.Fixed, QSizePolicy.Fixed)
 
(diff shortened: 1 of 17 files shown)


src/pymmcore_widgets/_group_preset_widget/_add_first_preset_widget.py
```diff
@@ -0,0 +1,171 @@
...
+            value = str(self.table.cellWidget(row, 1).value())
+            dev_prop_val.append((dev, prop, value))
+
+        self._preset = self.preset_name_lineedit.text()
+
+        with block_core(self._mmc.events):
+            for d, p, v in dev_prop_val:
+                self._mmc.defineConfig(self._group, self._preset, d, p, v)
+
+        self._mmc.events.configDefined.emit(self._group, self._preset, d, p, v)
```
tlambert03, line 151: when playing around with the example a bit:
1. I clicked the plus button
2. I gave a name to the group and clicked "create new group" (without adding any properties
3. I clicked preset ... and got an error:


```pytb
  File "/Users/talley/dev/self/pymmcore-widgets/src/pymmcore_widgets/_group_preset_widget/_add_first_preset_widget.py", line 151, in _create_first_preset
    self._mmc.events.configDefined.emit(self._group, self._preset, d, p, v)
UnboundLocalError: local variable 'd' referenced before assignment
```

so the user needs to somehow be prevented from getting to that point

tlambert03 (commented):
pretty fantastic functionality overall! 🚀


some general observations:

- Clicking either the Group -> "Edit" or Preset ->  "Edit" buttons when nothing is selected does nothing... which could be a bit confusing.  Those buttons should be disabled if nothing is selected
- similarly, If you click Preset -> "+" or "-" or Group "-" with nothing selected, nothing happens.  so it looks like the _only_ button that should be enabled if nothing is selected is the Group "+" button
- If you select a group, and begin to add a preset with Preset -> "+", you can of course add a new preset without giving it a name (it will give it an automatic `_n` suffix).  I think that's a bit confusing, since the user can't see what it's going to be named until they actually do it.  You could either set [`placeholderText`](https://doc.qt.io/qt-6/qlineedit.html#placeholderText-prop) on the line edit widget to what it would be if they don't enter anything... or raise/show an error message if they try to create a new one without entering anything (similarly to how you do with the new group window) ... but i don't think it should silently succeed without showing them what the name will be
- since there is a "save config" button I wonder if there should be a "load config" button right next to it?
- I wonder if it might be a bit _too_ easy to delete groups/presets with the "-" buttons?  Does micromanager just kill them without confirmation too?

--------

A general thought (happy to consider this in a later PR): I know this sequence of windows is what micromanager does, but I dislike the fact that you can have _three_ windows open: the main window, the Edit Group window, _and_ the Edit Preset window:

Screen Shot 2022-10-23 at 10 39 41 AM

I think three is too many popups at one time.  So we should think about how we might be able to use, e.g. the [`Sheet` window type](https://doc.qt.io/qt-6/qt.html#WindowType-enum), or generally compose a window that lets you edit _both_ groups and presets in the same window, rather than the explosion of popups

fdrgsp:
@tlambert03 

Thanks! I updated the code with your suggestions.

>  wonder if it might be a bit too easy to delete groups/presets with the "-" buttons? Does micromanager just kill them without confirmation too?

In this case I used a `QMessageBox` so that deleting happens only after pressing on the `Ok` button (as in micromanager).

___
For the issue of "too many windows opening" I agree as well but I think we can change it in a new PR, is that ok?

tlambert03:
almost there!! 

I think the only thing is still this:

> ... the user can't see what it's going to be named until they actually do it. You could either set [placeholderText](https://doc.qt.io/qt-6/qlineedit.html#placeholderText-prop) on the line edit widget to what it would be if they don't enter anything...

I can see that you did set the placeholder text if they haven't entered anything (which is great!) but that text is hard-coded to be `"NewPreset"` ... not what it _would_ be if they hit the enter button without editing the text (which might be, e.g., `"NewPreset_1"` etc...).  So, the placeholder text needs to properly reflect what the name would be if they don't enter anything

fdrgsp:
> I can see that you did set the placeholder text if they haven't entered anything (which is great!) but that text is hard-coded to be "NewPreset" ... not what it would be if they hit the enter button without editing the text (which might be, e.g., "NewPreset_1" etc...). So, the placeholder text needs to properly reflect what the name would be if they don't enter anything

@tlambert03 True! I meant to do that but then I forgot. I just added this functionality.

tlambert03 (approved):
big one here! :) LGTM

PR #50, opened 2022-09-23
fix: fix state error in snap and live buttons

tlambert03:
both the snap and live buttons have a bug where the internal state (`self._camera`) can go stale if the user uses `setCameraDevice`.  This PR removes the internal pointer to the device label, and always uses the core as the source of truth.

in general, I guess this is  reminder that we should, whenever possible, avoiding saving _any_ state on these widgets that has the potential to go out of sync with core (unless we _must_ save it for some reason)


ianhi (approved):
looks good!


A follow up (or here) we should try to deactivate if the camera device is ever set to `''` or if the only camera device is unloaded.

PR #58, opened 2022-10-13
fix: add vispy to pyproject.toml test

fdrgsp:
fix pre-test failing

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,11 +36,12 @@ dependencies = [
 # extras
 # https://peps.python.org/pep-0621/#dependencies-optional-dependencies
 [project.optional-dependencies]
-test = ["pytest>=6.0", "pytest-cov", "pytest-qt", "vispy"]
+test = ["pytest>=6.0", "pytest-cov", "pytest-qt"]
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -36,12 +36,11 @@ dependencies = [
 # extras
 # https://peps.python.org/pep-0621/#dependencies-optional-dependencies
 [project.optional-dependencies]
-test = ["pytest>=6.0", "pytest-cov", "pytest-qt"]
+test = ["pytest>=6.0", "pytest-cov", "pytest-qt", "vispy"]
 pyqt5 = ["PyQt5"]
 pyside2 = ["PySide2"]
 pyqt6 = ["PyQt6"]
 pyside6 = ["PySide6"]
-image = ['vispy']
```
ianhi, line 44: Still need this line for user installs so cannot remove it.

Given that the other optional dependencies work in the test my guess is that this wasn't working because it has single rather than double quotes.

ianhi (changes requested):

tlambert03 (approved):

PR #69, opened 2022-11-21
fix: fix sequence acquisition signal names

fdrgsp:
This PR replaces the old sequence acquisition signals names. To be merged **AFTER** https://github.com/pymmcore-plus/pymmcore-plus/pull/163.


tlambert03 (commented):
this will need a version pin as well

PR #70, opened 2022-11-23
fix: core -> mmcore

fdrgsp:
This PR renames all the `CMMCorePlus` widgets arguments as `mmcore`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,8 +16,8 @@ from pymmcore_plus import CMMCorePlus
 from pymmcore_widgets import StageWidget  # as an example... see below
 from qtpy.QtWidgets import QApplication
 
-core = CMMCorePlus.instance()
-core.loadSystemConfiguration()
+mmcore = CMMCorePlus.instance()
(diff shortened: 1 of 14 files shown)


tlambert03:
nice!  

note, the one test fail on pyside6 looks legitimate, but it's not part of the this PR, so we can fix elsewhere:

```
self = <pymmcore_widgets._pixel_size_widget.PixelSizeTable(0x2f6eeba4120) at 0x000002F6C64D2FC0>
mmcore = <CMMCorePlus at 0x2f6cb133d30>, parent = None

    def __init__(
        self, mmcore: CMMCorePlus, parent: Optional[QtW.QWidget] = None
    ) -> None:
        super().__init__(parent)
        self._mmc = mmcore
        hh = self.horizontalHeader()
>       hh.setSectionResizeMode(hh.Stretch)
E       AttributeError: 'PySide6.QtWidgets.QHeaderView' object has no attribute 'Stretch'
```

PR #71, opened 2022-11-23
fix: remove _update_mda_engine

fdrgsp:
This PR removes the `_update_mda_engine` method from `mda` and `explorer` since the new `pymmcore-plus PMDAEngine` does not have `events` anymore.

### DIFF
--- a/src/pymmcore_widgets/_mda_widget/_mda_widget.py
+++ b/src/pymmcore_widgets/_mda_widget/_mda_widget.py
@@ -16,16 +16,21 @@ from ._grid_widget import GridWidget
 from ._mda_gui import _MDAWidgetGui
 
 if TYPE_CHECKING:
-    from pymmcore_plus.mda import PMDAEngine
+    pass
 
 
 class MDAWidget(_MDAWidgetGui):
-    """Multi-dimensional acquisition Widget.
+    """A Multi-dimensional acquisition Widget.
 
-    The `MDAWidget` provides a GUI to construct a `useq.MDASequence` object.
+    The `MDAWidget` provides a GUI to construct a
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -39,7 +39,7 @@ test = ["pytest>=6.0", "pytest-cov", "pytest-qt"]
 pyqt5 = ["PyQt5"]
 pyside2 = ["PySide2"]
 pyqt6 = ["PyQt6"]
-pyside6 = ["PySide6!=6.4.0,!=6.4.0.1"]
```
tlambert03, line 42: do you know why?

tlambert03: locally... I get `AttributeError: 'PySide6.QtWidgets.QHeaderView' object has no attribute 'Stretch'`, which is actually a bug in the `GroupPresetTableWidget`.  You need to use the full name: `QHeaderView.ResizeMode.Stretch`

tlambert03: ...actually, I think we can even unpin version 6.4.0, that ultimately was `qtpy` not being ready for 6.4.0, but it works now

pyproject.toml
```diff
@@ -39,7 +39,7 @@ test = ["pytest>=6.0", "pytest-cov", "pytest-qt"]
...
+pyside6 = ["PySide6!=6.4.0,!=6.4.0.1,!=6.4.1"]
```
tlambert03, line 42:
```suggestion
pyside6 = ["PySide6"]
```

(and then apply fix https://github.com/pymmcore-plus/pymmcore-widgets/pull/71/files#r1030902381)

pyproject.toml
fdrgsp: ok I see, thanks!

tlambert03 (approved):
looks great!

PR #78, opened 2022-11-26
feat: General Widgets for MDAs

### LINKED ISSUE #76: task: remove all duplicated code between sample explorer and mda
@fdrgsp  ... I just noticed that `_mda_gui.py`/`_sample_explorer_gui.py`  and `_mda_widget.py`/`_sample_explorer_widget.py` are like 90% the same code :joy:

that definitely needs to be fixed... that's a maintenance problem waiting to happen.  I would suggest creating individual widgets for all of the duplicated code.  such as a `ZRange` widget that lets you select a zrange, a Channels table, a positions table, etc...  and then use those widgets in both the MDA and the sample explorer.

fdrgsp:
This PR removes duplicated code between mda and sample explorer.

It creates single widgets for channels, z-stacks, time and positions.

Closes #76

### DIFF
--- a/src/pymmcore_widgets/_general_mda_widgets.py
+++ b/src/pymmcore_widgets/_general_mda_widgets.py
@@ -1,0 +1,478 @@
+from typing import List, Optional
+
+from fonticon_mdi6 import MDI6
+from qtpy.QtCore import QSize, Qt
+from qtpy.QtWidgets import (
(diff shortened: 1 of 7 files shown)


tlambert03 (approved):
good stuff, thanks!

I'll probably follow up with some more dedup and refactoring, but this is way better.  thank you

PR #79, opened 2022-11-26
refactor: cleanup z tab selector

tlambert03:
I'd like us to continue the cleanup of the MDA/sample-explorer widgets from #78

This PR is an example of the kind of granularity I imagine.  It turns the `_MDAStackWidget` into a general `ZStackWidget` that has a `value` method that returns a dict in [useq-schema Z Plan format](https://pymmcore-plus.github.io/useq-schema/schema/axes/#z-plans)... this takes all the duplicated logic off of the mda widget and the sample explorer in their `get_state` and `set_state` methods.

Further, each tab in the widget implements a simple interface, with a `valueChanged` signal, a `value()` method, and a `z_range()` method, so most of the "mode-specific" logic can come off of the `ZStackWidget` itself.

I'd like us to do similar stuff with the other components in the mda and sample explorers.  For instance, the `_MDAPositionTable` could output a `list-of-dicts` in the `Tuple[Position, ...]` format expected by useq-schema, and all that duplicated logic could come off mda and sample explorers


tlambert03:
I didn't move any of the tests in this PR, but all of the tests in mda and sample explorer assert specific things about getting/setting z information could be in more granular unit tests now

fdrgsp (approved):
That's nice!

PR #80, opened 2022-11-27
fix: disable z checkbox at startup

fdrgsp:
minor fix. This PR unckeck the z-stack `QGroupWidget` checkbox at startup.

### DIFF
--- a/src/pymmcore_widgets/_mda_widget/_mda_gui.py
+++ b/src/pymmcore_widgets/_mda_widget/_mda_gui.py
@@ -60,6 +60,7 @@ class _MDAWidgetGui(QWidget):
         wdg_layout.addWidget(self.time_groupbox)
 
         self.stack_groupbox = ZStackWidget()
+        self.stack_groupbox.setChecked(False)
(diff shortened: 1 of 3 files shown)


tlambert03:
ah right ... I forgot to fix that.  That shouldn't be done in the `ZWidgetItself`, it should be done in whatever widget "uses" it (like the mda widget).  Think of it from the perspective of the z widget itself "knows".  It feels like it makes a little less sense if someone is making a Zwidget that it should always be disabled unless they enable it, (whereas it makes perfect sense for the mda widget to actively disable it, because it has additional context)

fdrgsp:
> ah right ... I forgot to fix that. That shouldn't be done in the `ZWidgetItself`, it should be done in whatever widget "uses" it (like the mda widget). Think of it from the perspective of the z widget itself "knows". It feels like it makes a little less sense if someone is making a Zwidget that it should always be disabled unless they enable it, (whereas it makes perfect sense for the mda widget to actively disable it, because it has additional context)

That's true, it make sense! I moved to the MDAs.

tlambert03 (approved):

PR #83, opened 2022-11-28
refactor: extract duplicated device filter widget code

tlambert03:
this includes, and should wait for #82


tlambert03:
after this PR, @fdrgsp could you see what the best way to remove the

PR #88, opened 2022-11-29
fix: minor fixes for PropertyBrowser and DeviceTypeFilters

fdrgsp:
- When first opening the `PropertyBrowser` widget, the `ReadOnly` properties are displayed but the `_read_only_checkbox` not checked. This PR fixes this issue.

- The `DeviceTypeFilters` signal `filtersChanged` does not have any arguments and connecting the `toggle` action of the `_read_only_checkbox` to  `self.filtersChanged.emit` won't work. Removing `.emit` solved the issue.

### DIFF
--- a/src/pymmcore_widgets/_property_browser.py
+++ b/src/pymmcore_widgets/_property_browser.py
@@ -2,6 +2,7 @@ from typing import Optional
 
 from pymmcore_plus import CMMCorePlus
 from qtpy.QtWidgets import QDialog, QHBoxLayout, QLineEdit, QVBoxLayout, QWidget
+from superqt.utils import signals_blocked
 
 from ._device_property_table import DevicePropertyTable
 from ._device_type_filter import DeviceTypeFilters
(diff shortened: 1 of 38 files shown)


fdrgsp:
> can you try `self._read_only_checkbox.toggled.connect(self.filtersChanged)` instead? (i.e. don't connect the `emit` function... connect the signal itself)

yes, that works!

src/pymmcore_widgets/_property_browser.py
```diff
@@ -29,6 +30,8 @@ def __init__(
 
         self._prop_table = DevicePropertyTable(mmcore=mmcore)
         self._device_filters = DeviceTypeFilters()
+        with signals_blocked(self._device_filters._read_only_checkbox):
+            self._device_filters._read_only_checkbox.setChecked(True)
```
tlambert03, line 34: this should just be `self._read_only_checkbox.setChecked(True)` around line 55 in the `DeviceTypeFilters` widget

tlambert03 (approved):

Issue #95, opened 2022-12-05
task: deduplicate sample explorer and mda


tlambert03:
hey @fdrgsp, after #94 gets in, I'd like for you to continue deduplicating the code between the mda and sample explorer.  

without looking line-for-line, it's not immediately obvious to me whether their methods are _straight_ duplications (where it would be as easy as just extracting them into a common subclass), or whether they began as a copy and paste, but have since acquired small changes that require going through it carefully.

as a relatively high priority, I'd like you to work on continuing in the vein of #79  and #94... extracting the logic for each subcomponent of the two widgets into a nicely self-contained object that can create one part of an MDA sequence.

The general goals to hit are:

- no code duplicated.  (if needed, you can use a common internal base class for methods that need to be shared between the two... for example `_update_total_time` on `MDAWidget` and on `SampleExplorerWidget` ... and all similar common methods)
- no accessing private attributes across widgets (this is a bit strict, but as an example, don't make a private widget named `self._lbl = QLabel()` in one widget, and then `other._lbl.setText('something')` in another widget.  Try to make a public method to set the state of the first widget
- no test duplications.  if two widgets add the same widget to their layouts, don't test the functionality of the inner widget in both of the outer widgets (e.g. deduplicate `test_gui_labels`  between `test_sample...` and `test_mda...`
- no two-names-for-the-same-thing.  i.e. don't do `self.short_name = self.much_longer_name`.  If you want to shorten the name _inside_ of the scope of a function, like `short = self.much_longer_name` ... that's ok, but don't assign it to self such that the widget now has two names pointing at the same thing
- make sure sample explorer also has a get_state and set_state like the mda widget
- each of the remaining subwidgets (channel and position, and more?) should have a `value()` method that returns a dict that would be suitable for passing to the `channels` and `positions` arguments in `MDASequence`.

(of course, this may be done over multiple PRs)

tlambert03:
sample explorer is gone now

PR #100, opened 2022-12-06
refactor: extract duplicated Channel Table code between mda and sample explorer

fdrgsp:
update MDAs Channel Table for task #95.


tlambert03:
looking good so far.  now need to remove the code in MDAWidget.set_state `# set channel table` around line 371.

since both sample and mda use the `_clear()` method.  make it public, and give it a docstring.
also, see if you can't make a method to replace the `#channel` related code in  `_update_total_time` in sample_explorer.

(basically... look around everywhere that this widget is being used – remembering that it also got renamed to `self.ch_gb` so you have to search for that too – and see if you can't avoid access _inner_ widgets of the channel table widget from other widgets)

src/pymmcore_widgets/_channel_table_widget.py
```diff
@@ -0,0 +1,228 @@
...
+            combo = cast(QComboBox, self.channel_tableWidget.cellWidget(row, 0))
+            used_idxs.append(combo.currentIndex())
+        new_idxs = list(set(idxs) - set(used_idxs))
+        return new_idxs[0] if new_idxs else 0
+
+    def _remove_channel(self) -> None:
+        rows = {r.row() for r in self.channel_tableWidget.selectedIndexes()}
+        for idx in sorted(rows, reverse=True):
+            self.channel_tableWidget.removeRow(idx)
+        self.valueChanged.emit()
```
tlambert03, line 174: all of these `valueChanged.emit` emissions rely on someone actually clicking a button, but that leaves the possibility of inconsistent states if something programmatically adjust the table.

Note also that the `_create_gui` method in *both* the MDA widget and Sample widget guis have these lines:

```python
self.channel_groupbox.channel_tableWidget.model().rowsInserted.connect(
    self._enable_run_btn
)
self.channel_groupbox.channel_tableWidget.model().rowsRemoved.connect(
    self._enable_run_btn
)
```

So, that needs work.  The `ChannelTable.valueWidget` should emit pretty much any time anything happens that would make the `value()` method return a different value (i.e. anytime the the table changes at all).  And other widgets should connect to _that_ signal, rather than accessing the inner table widget.  That is, it should be `self.channel_groupbox.valueChanged.connect...` rather than `self.channel_groupbox.channel_tableWidget.model().rowsInserted.connect...`

PR #105, opened 2022-12-09
feat: add core signals to ChannelTable widget

fdrgsp:
clear channel `QTableWidget` on `systemConfigurationLoaded` + add configGroupDeleted signal.


ianhi:
Should also try to update the `channelGroup`

src/pymmcore_widgets/_mda/_channel_table_widget.py
```diff
@@ -109,6 +109,19 @@ def __init__(
 
         group_layout.addWidget(wdg, 0, 1)
 
+        self._mmc.events.systemConfigurationLoaded.connect(self._clear)
```
tlambert03, line 112: did you mean to connect `_clear` or `_on_sys_cfg_loaded`?    if the former, update the disconnect, if the latter, update here

PR #109, opened 2022-12-13
refactor: remove duplicate GridParam widget

tlambert03:
@fdrgsp, can you review this one?

Not totally sure whether there was anything different between the grid widget in the sample explorer module, and the one in the _grid_widget module

### DIFF
--- a/src/pymmcore_widgets/_mda/_grid_widget.py
+++ b/src/pymmcore_widgets/_mda/_grid_widget.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+from typing import cast
+
 from pymmcore_plus import CMMCorePlus
 from qtpy.QtCore import Qt, Signal
 from qtpy.QtWidgets import (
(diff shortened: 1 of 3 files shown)


fdrgsp:
@tlambert03 I've been thinking that at this point mda and explorer are quite similar in functionality (when I first created the explorer, the mda could not do grid acquisition, I added that later on) it might be good to completely remove the sample explorer widget 😂! It is only different in napari-micromanager in the way how the acquire grid are displayed so maybe we can find a way to do that with only the mda. What do you think?

tlambert03:
lol :joy: ... well if you feel like you can accomplish what you want with the "grid" button in the position list of the MDA widget, that would be fine with me :)  but i'll let you make the call

fdrgsp:
> lol 😂 ... well if you feel like you can accomplish what you want with the "grid" button in the position list of the MDA widget, that would be fine with me :) but i'll let you make the call

Great! When #105, #110, #111 and #112 will be merged, I think will open a PR where I remove the explorer!

tlambert03:
closing, I think this is superseded by some of the changes made in the position refactors

PR #110, opened 2022-12-13
fix: fix typing in sample explorer

tlambert03:
cc @fdrgsp 

using `useq.Position` was a good way to get around the typing issues in SampleExlorer (the `tuple[str, float, float, float] | tuple[str, float, float]` issue)... particularly since that was the ultimate desired output type anyway.

An alternative would have also been to use a `typing.NamedTuple`, which would let you have an optional fourth "z" parameter (so you don't need to deal with the 3-tuple or 4-tuple everywhere)

### DIFF
--- a/src/pymmcore_widgets/_mda/_sample_explorer_widget.py
+++ b/src/pymmcore_widgets/_mda/_sample_explorer_widget.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import warnings
-from typing import cast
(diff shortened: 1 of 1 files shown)


tlambert03:
I also moved all of the stuff that *wasn't* changing outside of the `for st_pos in explorer_starting_positions:` loop in the `_create_grid_coords` function.  As a general rule, it's a good idea to leave all of the calculations that aren't actually sensitive to the thing you're iterating over above the for loop.

PR #111, opened 2022-12-16
fix: fix MDA run_button not showing when _include_run_button=True

fdrgsp:
The MDA `run_button` is not showing when `_include_run_button` is `True`. By mistake this has been changed and this PR fixes it.


tlambert03 (approved):
Looks good to me, if you're so inclined, anytime you find something like this, a small test will help ensure that there is no regression in the future :)

PR #113, opened 2022-12-17
feat: updated version of Channel Table Widget

### LINKED ISSUE #125: unable to select channel in mda widget if is widget loaded before cfg
not sure if this is widgets or napari-micromanager, but i suspect widgets:

if I open the mda widget before loading a system config, I seem to be never able to add a channel later, even after loading a config:

https://user-images.githubusercontent.com/1609449/220792317-39c0a121-00ec-448a-a6d9-57c770e3ebec.mov

if I do it in the reverse order, it's fine:

https://user-images.githubusercontent.com/1609449/220792345-97662413-3d7f-4bff-a0f4-73f06b3dbb47.mov

fdrgsp:
This PR mainly:
 - updates the current `ChannelTable` widget by adding a `QComboBox` that can be used to select which micromanager `ChannelGroup` to use in the `ChannelTable` (video below).

- adds a general `ChannelGroupWidget` which is a simple `QComboBox` that can be used  to keep track/change the micromanager `ChannelGroup` property.

- adds `ChannelTable` widget and `ChannelGroupWidget` to the docs.

- adds new tests for both.

merge after https://github.com/pymmcore-plus/pymmcore-plus/pull/180.
___

https://user-images.githubusercontent.com/70725613/213214341-73d4989d-8907-4ffb-9e9f-2037a862ec5f.mov

fixes #125


src/pymmcore_widgets/_mda/_channel_table_widget.py
```diff
@@ -109,22 +122,53 @@ def __init__(
...
+            items = [combo.itemText(ch) for ch in range(combo.count())]
+            if group == combo.whatsThis() and config in items:
+                combo.clear()
+                combo.addItems(self._mmc.getAvailableConfigs(group))
+                if self._mmc.getChannelGroup() != config:
+                    combo.setCurrentText(self._mmc.getChannelGroup())
 
     def channelGroup(self) -> str:
         """Return current channel group."""
         return self._channel_group
```
tlambert03, line 153: does this attribute even exist anymore?

tlambert03 (commented):
i guess my main question here would be: haven't we lost some flexibility?

let's say someone had a configuration that had _two_ config groups that were meaningful as channel groups... and they wanted to have a channel table widget dedicated to each of them.  Wouldn't this prevent that, since now the channel table widget is _obligated_ to follow the current group?

also, i think there are some latent bugs here, since `_channel_group` is accessed in places but isn't actually set anywhere.  I'm going to request that this have complete test coverage before merge

src/pymmcore_widgets/_mda/_channel_table_widget.py
fdrgsp: > let's say someone had a configuration that had _two_ config groups that were meaningful as channel groups... and they wanted to have a channel table widget dedicated to each of them. Wouldn't this prevent that, since now the channel table widget is _obligated_ to follow the current group?

Yes, that's true! What about if we keep both versions? one that is more general (the current one, which we can call something like `GeneralChannelTable`) and this one (`ChannelTable`) which is more similar to micromanager...or we can implement it only in `napari-micromanager` if you think it is better!

> also, i think there are some latent bugs here, since `_channel_group` is accessed in places but isn't actually set anywhere.

That's true, I forgot to remove the variable in few places...Also at the moment the micromanager `ChannelGroup` property is not controlled by the `ChannelTable` widget itself but is only handled by the `ChannelGroupWidget` that I have added in this PR (`_channel_group_widget.py`).

About testing this widget, I added some tests to check that everything worked as I wanted - `test_channel_table_widget.py`.

src/pymmcore_widgets/_channel_group_widget.py
```diff
@@ -0,0 +1,94 @@
...
+    mmcore : CMMCorePlus | None
+        Optional [`pymmcore_plus.CMMCorePlus`][] micromanager core.
+        By default, None. If not specified, the widget will use the active
+        (or create a new)
+        [`CMMCorePlus.instance`][pymmcore_plus.core._mmcore_plus.CMMCorePlus.instance].
+    """
+
+    def __init__(
+        self,
+        connect_core: bool = True,
```
tlambert03, line 27: make keyword only

src/pymmcore_widgets/_mda/_channel_table_widget.py
```diff
@@ -34,32 +36,43 @@ class ChannelDict(TypedDict, total=False):
...
 
     The `value()` method returns a dictionary with the current state of the widget, in a
     format that matches one of the [useq-schema Channel
     specifications](https://pymmcore-plus.github.io/useq-schema/schema/axes/#useq.Channel).
+
+    Parameters
+    ----------
+    title : str
+        Title of the QGroupBox widget. Bt default, 'Channel'.
+    connect_core : bool
```
tlambert03, line 50: move kwarg only

src/pymmcore_widgets/_mda/_channel_table_widget.py
```diff
@@ -34,32 +36,43 @@ class ChannelDict(TypedDict, total=False):
...
         parent: QWidget | None = None,
         *,
-        channel_group: str = "",
         mmcore: CMMCorePlus | None = None,
     ) -> None:
         super().__init__(title, parent=parent)
 
         self._mmc = mmcore or CMMCorePlus.instance()
-        self.setChannelGroup(channel_group or self._mmc.getChannelGroup())
+        self._connect_core = connect_core
```
tlambert03, line 75: remove if not used internally

tlambert03:
can you update the original description here to clarify the motivation for this pr?  could be a bullet point list of what problem it solves, what wasn't possible  before what is possible now.

src/pymmcore_widgets/_mda/_channel_table_widget.py
```diff
@@ -34,32 +34,43 @@ class ChannelDict(TypedDict, total=False):
...
+    mmcore : CMMCorePlus | None
+        Optional [`pymmcore_plus.CMMCorePlus`][] micromanager core.
+        By default, None. If not specified, the widget will use the active
+        (or create a new)
+        [`CMMCorePlus.instance`][pymmcore_plus.core._mmcore_plus.CMMCorePlus.instance].
     """
 
     valueChanged = Signal()
-    _channel_group: str
+    CH_GROUP_ROLE = 1
```
tlambert03, line 61: 1 is reserved for DecorationRole... see https://doc.qt.io/qt-6/qt.html#ItemDataRole-enum

use a UserRole plus some offset

tlambert03 (approved):
good to go after changing the data role number

src/pymmcore_widgets/_mda/_channel_table_widget.py
fdrgsp: is it ok to do that?

`CH_GROUP_ROLE = Qt.UserRole + 1`

and then to set the item data:

`channel_combo.setItemData(0, _channel_group, self.CH_GROUP_ROLE)`

The index depends on the number of items in the combobox...

PR #119, opened 2023-01-11
refactor: extract basic position table

fdrgsp:
This PR create MDAs `PositionTable` + test for task #95 in line with #79, #94 and #100.

Screen Shot 2023-01-10 at 11 17 30 PM


src/pymmcore_widgets/_mda/_mda_widget.py
```diff
@@ -425,6 +251,20 @@ def get_state(self) -> MDASequence:
             time_plan=time_plan,
         )
 
+    def _get_current_position(self) -> list[PositionDict]:
+        return [
+            {
+                "name": "Pos000",
```
tlambert03, line 257: quick note here (this can come elsewhere though):
whenever you create a string pattern as you've done here with `Grid{some_variable}_Pos{some_variable}` that you intend to decode somewhere else, you should make a single place where that "code" is defined.  That could be something like a string template:
```python
GRID_TEMPLATE = 'Grid{grid}_Pos{pos}'
```
... that you later format: `GRID_TEMPLATE.format(grid=..., pos=...)`

or it could be a function:

```python
def encode_grid(grid: int, pos: int) -> str:
    return f'Grid{grid}_Pos{pos}'
```

that has the benefits of 1) making it clear that "this is an encoding strategy here" ... i.e that the "Grid" and "Pos" parts in there are actually important somewhere else and 2) makes it trivial to change later without having to look for all the places you used the `Pos___` literal

tlambert03 (approved):

PR #121, opened 2023-01-19
fix: Stop shutter widget from accidentally closing shutter

ianhi:
The shutter widget should not be so eagerly modifying shutter state. It was closing all shutters it interacted with on initialization and also interfering with autoshutter behavior by closing all shutters after snapping an image. This is best left to control by the core AutoShutter mechanism. Particularly a problem as for some scopes the BF light source is defined as shutter device, so the widget was turning off a halogen lamp which could never warm up in time for a fast exposure.

This was causing a bug for @JasonYu1 and me where all our BF images had no signal. Some test code demonstrating this:

```python
import numpy as np
from pymmcore_plus import CMMCorePlus, DeviceType
from pymmcore_widgets import ShuttersWidget


# need a qapp for the widgets to be created
app = QApplication([])

core = CMMCorePlus.instance()
core.loadSystemConfiguration("new-camera-chan.cfg")

core.setShutterOpen("Lamp", True)

# I have three shutter devices
print(core.getLoadedDevicesOfType(DeviceType.ShutterDevice))

core.setConfig("Filter", "BF")
core.setExposure(500)

# the default shutter for autoshutter should now
# be the BF Shutter
print(core.getShutterDevice())

# expect large numbers as lamp is on
img = core.snap()
print(f"image mean: {img.mean()}")


# now add a shutter widget
lamp_shutter_wdg = ShuttersWidget("Lamp")


# The shutter widget turned the lamp off!
img = core.snap()
print(f"image mean: {img.mean()}")
```

Before this fix this results in:
```
('Lamp', 'BF Shutter', 'Fluor-Shutter')
BF Shutter
image mean: 24546.016795199524
image mean: 468.0492142022825
```

note that the second image has essentially no intensity (and also I can see that the light never hits the sample)

and after the fix gives:

```
('Lamp', 'BF Shutter', 'Fluor-Shutter')
BF Shutter
image mean: 24569.787509095437
image mean: 24581.002033787147
```

(and I can see the light turning on for both exposures.

### DIFF
--- a/src/pymmcore_widgets/_shutter_widget.py
+++ b/src/pymmcore_widgets/_shutter_widget.py
@@ -73,6 +73,9 @@ class ShuttersWidget(QWidget):
         self._mmc.events.autoShutterSet.connect(self._on_autoshutter_changed)
         self._mmc.events.propertyChanged.connect(self._on_prop_changed)
         self._mmc.events.configSet.connect(self._on_channel_set)
+        self._mmc.events.continuousSequenceAcquisitionStarted.connect(self._on_live_mode)
+        self._mmc.events.sequenceAcquisitionStarted.connect(self._on_live_mode)
+        self._mmc.events.sequenceAcquisitionStopped.connect(self._on_live_mode)
 
         self.destroyed.connect(self._disconnect)
 
@@ -273,6 +276,10 @@ class ShuttersWidget(QWidget):
                 if value != "Undefined":
                     self._mmc.events.propertyChanged.emit(value, "State", True)
 
+    def _on_live_mode(self) -> None:
+        if not self._mmc.getShutterOpen(self.shutter_device):
+            self._set_shutter_wdg_to_closed()
+
     def _on_channel_set(self, group: str, preset: str) -> None:
         if (self._mmc.getShutterDevice() == self.shutter_device) and self._mmc.getAutoShutter():
             self.shutter_button.setEnabled(False)
@@ -300,8 +307,6 @@ class ShuttersWidget(QWidget):
                 self.autoshutter_checkbox.setChecked(state)
         if self._mmc.getShutterDevice() == self.shutter_device:
             self.shutter_button.setEnabled(not state)
-        if state and self._mmc.isSequenceRunning():
-            self._mmc.stopSequenceAcquisition()
 
     def _close_shutter(self, shutter: str) -> None:
         self._set_shutter_wdg_to_closed()
--- a/src/pymmcore_widgets/_shutter_widget.py
+++ b/src/pymmcore_widgets/_shutter_widget.py
@@ -243,7 +243,10 @@ class ShuttersWidget(QWidget):
                 self.autoshutter_checkbox.setChecked(self._mmc.getAutoShutter())
             else:
                 self.autoshutter_checkbox.setChecked(False)
-            self.shutter_button.setEnabled(True)
+            if self._mmc.getShutterDevice() == self.shutter_device:
+                self.shutter_button.setEnabled(not self._mmc.getAutoShutter())
+            else:
+                self.shutter_button.setEnabled(True)
             if self._mmc.getShutterOpen(self.shutter_device):
                 self._set_shutter_wdg_to_opened()
             else:
--- a/src/pymmcore_widgets/_shutter_widget.py
+++ b/src/pymmcore_widgets/_shutter_widget.py
@@ -72,12 +72,6 @@ class ShuttersWidget(QWidget):
         self._mmc.events.systemConfigurationLoaded.connect(self._refresh_shutter_widget)
         self._mmc.events.autoShutterSet.connect(self._on_autoshutter_changed)
         self._mmc.events.propertyChanged.connect(self._on_prop_changed)
-        self._mmc.events.continuousSequenceAcquisitionStarted.connect(
-            self._on_seq_started
-        )
-        self._mmc.events.sequenceAcquisitionStarted.connect(self._on_seq_started)
-        self._mmc.events.sequenceAcquisitionStopped.connect(self._on_seq_stopped)
-        self._mmc.events.imageSnapped.connect(self._on_seq_stopped)
         self._mmc.events.configSet.connect(self._on_channel_set)
 
         self.destroyed.connect(self._disconnect)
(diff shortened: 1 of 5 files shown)


ianhi:
TODO:

1. Update the tests
2. Push my local changes continuing to clean up enabled vs not enabled
3. Make sure that the `snap` widget also pushes through a State Closed after a snap
     - https://github.com/pymmcore-plus/pymmcore-widgets/blob/59074e005096b138c6e359cbce9b768b1135ba6e/src/pymmcore_widgets/_snap_button_widget.py#L82-L86

src/pymmcore_widgets/_shutter_widget.py
```diff
@@ -244,25 +238,21 @@ def _refresh_shutter_widget(self) -> None:
...
             props = self._mmc.getDevicePropertyNames(self.shutter_device)
             self._is_multiShutter = bool([x for x in props if "Physical Shutter" in x])
 
-    def _on_seq_started(self) -> None:
-        if self._mmc.getShutterOpen(self.shutter_device):
-            self._set_shutter_wdg_to_opened()
-
-    def _on_seq_stopped(self) -> None:
-        self._close_shutter(self.shutter_device)
-
```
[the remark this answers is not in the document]
ianhi: I agree that it's really nice to be able to have a visual indication of the state of a continuous acquisition. I just don't thnk t that the shutter way to represent that because depending on config file/autoshutter state the shutter state may or may not be tied to the sequence acquisition.  And then it's very problematic for this widget to be potentially modifying the state to be not what the user wanted.


So I think that something like you proposed:

> .maybe a good thing to do can be to modify the _on_seq_stopped method so that it won't close the shutter but only update its state (icon, etc..)

Could work really well - because unfortunately not all devices send the correct signals from the device adapter, so this is a good time to check if it's closed.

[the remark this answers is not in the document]
tlambert03: > And then it's very problematic for this widget to be potentially modifying the state to be not what the user wanted.

This makes sense,  I haven't really been following yet here, but a general thought: widgets should be views onto a model (core), and controllers that accept user input. But nothing more.  They should modify any state that doesn't directly come from user interaction... and should otherwise attempt to show (view) the state of the core

Is that happening here? Or is it an interesting/special case?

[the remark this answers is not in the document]
ianhi: > but a general thought: widgets should be views onto a model (core), and controllers that accept user input. But nothing more.

100% agree! I think a nice test to add would be to:

1. Grab the core state
2. create every widget
3. check that the state is unmodified


> Is that happening here? Or is it an interesting/special case?

The goal of this PR is to move to view on the core approach. Previously there were several ways in which this widget would close shutters when it ought not to. I think that this was an approach to ensure that they stayed in sync with the core state as unfortunately the signalling from core that the shutter opened or closed is not perfect for many devices. 

I think the special cases being discussed are more so extra times that we should query the true state of the shutter - e.g. at the end of an acquisition is a good time  to check if the shutter has closed.

ianhi:
I'm not abandoning this, but may be a week until I can properly return - currently traveling and away from a physical microscope to to check everything on.

ianhi:
I have returned! and despite the fog of jet lag I think I've now got everything working as I'd expect. Presuming tests pass here (they do locally) then I think this is ready for review.

tlambert03 (approved):
Fine with me, will give it a day or two for @fdrgsp to have a chance to comment, then will merge

src/pymmcore_widgets/_shutter_widget.py
```diff
@@ -280,16 +278,17 @@ def _on_prop_changed(self, dev_name: str, prop_name: str, value: Any) -> None:
                 if value != "Undefined":
                     self._mmc.events.propertyChanged.emit(value, "State", True)
 
+    def _on_live_mode(self) -> None:
+        if not self._mmc.getShutterOpen(self.shutter_device):
+            self._set_shutter_wdg_to_closed()
+
```
ianhi, line 284: @fdrgsp note that I added in the check you suggested whenever on sequence acquiistion starts or stops.

ianhi: ah, and also see that it's not complete....

ianhi: now fixed for real

src/pymmcore_widgets/_shutter_widget.py
```diff
@@ -244,25 +243,24 @@ def _refresh_shutter_widget(self) -> None:
...
                 self.autoshutter_checkbox.setEnabled(True)
-                self.autoshutter_checkbox.setChecked(True)
-                self.shutter_button.setEnabled(False)
+                self.autoshutter_checkbox.setChecked(self._mmc.getAutoShutter())
             else:
+                self.autoshutter_checkbox.setChecked(False)
+            if self._mmc.getShutterDevice() == self.shutter_device:
                 self.shutter_button.setEnabled(not self._mmc.getAutoShutter())
+            else:
+                self.shutter_button.setEnabled(True)
```
fdrgsp, line 254: @ianhi why do we need to do this comparison? isn't it better to just do

```python
self.shutter_button.setEnabled(not self._mmc.getAutoShutter())
```

since when the `autoshutter` is on all the shutter should be disabled? 

For example, if you run this code, the shutter s2 is enabled even if it should not...

```python
from pymmcore_widgets import ShuttersWidget
mmc = CMMCorePlus.instance()
mmc.loadSystemConfiguration()
s1 = ShuttersWidget("White Light Shutter")
s2 = ShuttersWidget("LED Shutter")
s1.show()
s2.show()
```

src/pymmcore_widgets/_shutter_widget.py
```diff
@@ -310,10 +311,8 @@ def _on_autoshutter_changed(self, state: bool) -> None:
         if self.autoshutter:
             with signals_blocked(self.autoshutter_checkbox):
                 self.autoshutter_checkbox.setChecked(state)
-        self.shutter_button.setEnabled(not state)
-
-        if state and self._mmc.isSequenceRunning():
-            self._mmc.stopSequenceAcquisition()
+        if self._mmc.getShutterDevice() == self.shutter_device:
+            self.shutter_button.setEnabled(not state)
```
fdrgsp, line 315: same here as above, should we just do this?
```python
self.shutter_button.setEnabled(not state)
```

src/pymmcore_widgets/_shutter_widget.py
ianhi: > since when the autoshutter is on all the shutter should be disabled?

I think this may not be true. To my understanding the way autoshutter works is the core will take control of the current shutter device (of which there can only ever be one) and toggle that open and closed when images are taken, while leaving other shutters in whatever state the user wants. 

So in your example I'd argue that while `s2` should indeed not be toggled open (given the naming scheme) there is no reason that the button should be disabled - i.e. if the user wanted to open it for some reason then we shouldn't prevent them from doing so.

src/pymmcore_widgets/_shutter_widget.py
ianhi: Same explanation as the other one, Autoshutter only controls one shutter at a time so only that shutter should be disabled.

src/pymmcore_widgets/_shutter_widget.py
fdrgsp: ok I see your point! :thumbsup:

ianhi:
thanks all! I will use the updated main branch on a real scope today for an experiment so this is very helpful

PR #126, opened 2023-03-03
feat: update GridWidget and PositionTable widget

fdrgsp:
This PR updates the `GridWidget` and the `PositionTable` widget to exploit the new `MDASequence` update (https://github.com/pymmcore-plus/useq-schema/pull/81 - `grid_plan` and `MDASequence` in `Position`).

NOTE: I've also fixed the `SampleExplorer` widget even if I think we can remove it after this PR is merged.
___
Screenshot 2023-03-03 at 1 55 56 PM

### DIFF
--- a/src/pymmcore_widgets/_mda/_grid_widget.py
+++ b/src/pymmcore_widgets/_mda/_grid_widget.py
@@ -557,15 +557,10 @@ class GridWidget(QDialog):
         """Set the state of the widget from a useq AnyGridPlan or dictionary."""
         grid_plan = MDASequence(grid_plan=grid).grid_plan
 
-        overlap = grid.get("overlap") or 0.0
-        over_x, over_y = (
-            overlap if isinstance(overlap, (tuple, list)) else (overlap, overlap)
-        )
+        over_x, over_y = grid_plan.overlap
         self.overlap_spinbox_x.setValue(over_x)
         self.overlap_spinbox_y.setValue(over_y)
-        ordermode = grid.get("mode") or OrderMode.row_wise_snake
-        ordermode = ordermode.value if isinstance(ordermode, OrderMode) else ordermode
-        self.ordermode_combo.setCurrentText(ordermode)
+        self.ordermode_combo.setCurrentText(grid_plan.mode.value)
 
         if isinstance(grid_plan, GridRelative):
             self.tab.setCurrentIndex(0)
--- a/src/pymmcore_widgets/_mda/_grid_widget.py
+++ b/src/pymmcore_widgets/_mda/_grid_widget.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import math
 from typing import TYPE_CHECKING, Literal
 
+from pydantic import ValidationError
 from pymmcore_plus import CMMCorePlus
 from qtpy.QtCore import Qt, Signal
 from qtpy.QtWidgets import (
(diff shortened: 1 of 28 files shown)


tlambert03:
v0.2.0 of useq schema should be released now

tlambert03:
just want to note somewhere:

I'm removing the calls to `set_fov_size` here.  my intention was for that to be called at the last minute by pymmcore-plus (or whatever is running the sequence) not the thing that set's the sequence.

src/pymmcore_widgets/_mda/_positions_table_widget.py
```diff
@@ -213,88 +267,217 @@ def _add_position_row(self) -> int:
...
-                # remove if is a single position
-                self._table.removeRow(idx)
+    def _get_grid_buttons(self, row: int) -> tuple[QPushButton, QPushButton]:
+        return (
+            self._table.cellWidget(row, 4).layout().itemAt(0).widget(),
+            self._table.cellWidget(row, 4).layout().itemAt(1).widget(),
+        )
+
+    def _grid_widget(self) -> None:
+        if not self._mmc.getXYStageDevice():
```
tlambert03, line 334: this feels a bit like a bleeding of concerns.  You are making it so that this widget can *never* be used to setup a grid plan when there isn't an active stage device.  But that feels like the kind of thing that an application should do (something like napari micromanager) not something that a library of widgets should be doing

src/pymmcore_widgets/_mda/_grid_widget.py
```diff
@@ -1,228 +1,607 @@
...
+
+        corners = self._create_corners_grid_wdg()
+        self.tab.addTab(corners, "Grid from Corners")
+
+        layout.addWidget(self.tab)
+
+        self.tab.currentChanged.connect(self._update_info)
+        return wdg
+
+    def _create_row_cols_wdg(self) -> QWidget:
```
tlambert03, line 207: i really don't like these methods that exist just to create an return specialized widget... instead, make a new class, and instantiate it:
```python
class _RowsColsWdg(QWidget):
    def __init__(self):
        self.setLayout(QGridLayout())
        ...  # etc
```

src/pymmcore_widgets/_mda/_positions_table_widget.py
```diff
@@ -213,88 +267,217 @@ def _add_position_row(self) -> int:
...
+            self._grid_wdg.close()  # type: ignore
 
-        self._rename_positions()
+        self._grid_wdg = GridWidget(
+            parent=self,
+            mmcore=self._mmc,
+            current_stage_pos=(self._mmc.getXPosition(), self._mmc.getYPosition()),
+        )
+        row = self._table.indexAt(self.sender().parent().pos()).row()
+        self._grid_wdg.valueChanged.connect(lambda x: self._add_grid_plan(x, row))
```
tlambert03, line 346: this is a bit of a circuitous way to use a Dialog (i.e. you're not taking advantage of what a QDialog does). 

You have a button ("add grid") that essentially "accepts" the dialog, but then emits a signal, and the _other_ widget is hooked up to that value emission and does the job of closing the dialog.

Modal QDialogs have a whole structure for doing exactly this (see discussion of `accept`  and `reject` in https://doc.qt.io/qt-6/qdialog.html#modal-dialogs ...

tlambert03:
given the size of this PR, there's a lot of patterns that could potentially be cleaned up and improved.   but that process could take a long time too... so feel free to address the couple comments I made, but since this touches a relatively "hidden" bit of the interface, I'm also ok with just getting this in and moving on

tlambert03:
@fdrgsp, needs more comments and docstrings.  For example, I find myself at the moment reading through the code of `_DoubleSpinboxWidget` trying to figure out what it does.  Ideally, every class and method would have at least a single sentence summarizing why it needs to exist.  Minimally, every class should have a quick one-liner

PR #129, opened 2023-03-10
feat: add "advanced" channel options to ChannelTable

fdrgsp:
This PR adds the possibility (enabled through a `QCheckBox`) to set the `z_offset`, `do_stack` and `acquire_every` options to every channel in the `ChannelTable`.

---
Screenshot 2023-03-10 at 9 32 51 AM


tlambert03:
awesome.  as discussed, in a follow up PR, we'll add:

- if the advanced button is *not* checked, and some of the advanced values are *not* their default values... then show a little warning icon to the right of the advanced checkbox that has info explaining that advanced values have been selected (on hover)

PR #128, opened 2023-03-07
[WIP] feat: add combo for micromanager Focus and Autofocus to PositionTable

fdrgsp:
merge after #140.

note: tests will fail until useq-schema is upgraded (I'm working on a PR for that)

This PR adds "control" for micromanager `Focus` and `AutoFocus` (e.g. Nikon PFS) devices.

This PR is linked to https://github.com/pymmcore-plus/useq-schema/issues/4 and https://github.com/pymmcore-plus/pymmcore-plus/issues/51#issuecomment-1397585287.

Screenshot 2023-05-23 at 3 57 54 PM

to test:
```python
from pymmcore_plus import CMMCorePlus
from pymmcore_widgets._mda import PositionTable
mmc = CMMCorePlus.instance()
cfg = "./tests/test_config.cfg"
mmc.loadSystemConfiguration(cfg)
p = PositionTable()
p.show()
```

PR #133, opened 2023-03-30
[WIP] feat: add autofocus control (ian)

fdrgsp:
@ianhi  link to [#128 ](https://github.com/pymmcore-plus/pymmcore-widgets/pull/128#issuecomment-1490590164)

This PR  adds "control" for micromanager Focus and AutoFocus (e.g. Nikon PFS) devices.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,6 +30,9 @@ jobs:
         python-version: ["3.8", "3.10"]
         backend: [pyside2, pyqt5]
         include:
+          - platform: macos-latest
+            python-version: "3.10"
+            backend: pyside6
           - platform: windows-latest
             python-version: "3.10"
             backend: pyside6
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,7 +20,7 @@ repos:
       - id: trailing-whitespace
 
   - repo: https://github.com/charliermarsh/ruff-pre-commit
-    rev: v0.0.260
+    rev: v0.0.263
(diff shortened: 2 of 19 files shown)


ianhi (changes requested):
The population of the combo boxes is great, but there are some issues with the logic afterwards. I'm not able to populate my `PFS-Offset` as it only every fills in `focus` also as i note in a line level comment the combo boxes should not be changing what the current focus device is, only modifying what happens when we click `add`.

Also some open questions to consider:

1. What should if we start adding positoins, and then modify the combo boxes
2. Should we be recording two values? If you are using autofocus the only value you need to define for the position table is the `PFS-offset` because you will be constatnyl updating the absolute position in order to correct for drift

fdrgsp:
> I'm not able to populate my `PFS-Offset` as it only every fills in `focus`

@ianhi Does it means that you see it only in the `Z Focus` combo and not also in the `Z AutoFocus` one? If that's the case it's weird because both the combo are populated with all the devices of type `DeviceType.StageDevice` and the `PSFOffset` is of this type on our scope...

> the combo boxes should not be changing what the current focus device is
Sure, you're right, I changed it and the stage info are stored in a dictionary. We can find a better way but now the combo are disconnected from the core.

> What should if we start adding positoins, and then modify the combo boxes

At the moment when you change the z combo, only the column that is selected is shown and the `add `button will add a value only in that column. The `value()` method will use the value of the "active" z column. What do you think?

fdrgsp:
continue in #147

PR #142, opened 2023-06-05
fix: set the sequence fov size when generating the sequence + test

fdrgsp:
This PR adds the FOV size (width, height) info to the `MDASequence` that we create through the `MDAWidget`. Without this info, the `grid plan` won't work because the FOV is considered to be (1, 1).

Should it be here? in pymmcore-plus?...not sure...

### DIFF
--- a/src/pymmcore_widgets/_mda/_mda_widget.py
+++ b/src/pymmcore_widgets/_mda/_mda_widget.py
@@ -235,6 +235,8 @@ class MDAWidget(QWidget):
             self.time_groupbox.value() if self.time_groupbox.isChecked() else None
         )
 
+        _, _, width, height = self._mmc.getROI()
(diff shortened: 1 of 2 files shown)


src/pymmcore_widgets/_mda/_mda_widget.py
```diff
@@ -243,20 +245,24 @@ def get_state(self) -> MDASequence:
...
 
-        return MDASequence(
+        sequence = MDASequence(
             axis_order=self.buttons_wdg.acquisition_order_comboBox.currentText(),
             channels=channels,
             stage_positions=stage_positions,
             z_plan=z_plan,
             time_plan=time_plan,
         )
+        sequence.set_fov_size(fov)
```
tlambert03, line 263: move this to pymmcore-plus prepare to run

tlambert03:
this will be moved to pymmcore-plus

PR #150, opened 2023-07-13
feat: save pixel size config in GroupPresetTableWidget + test

fdrgsp:
When calling the core method `saveSystemConfiguration`, the pixel size info are not saved. This PR updates the `GroupPresetTableWidget` `_save_cfg` method so that also the pixel size configurations are saved in the .cfg file.

### DIFF
--- a/src/pymmcore_widgets/_group_preset_widget/_group_preset_table_widget.py
+++ b/src/pymmcore_widgets/_group_preset_widget/_group_preset_table_widget.py
@@ -401,7 +401,26 @@ class GroupPresetTableWidget(QGroupBox):
             self, "Save Micro-Manager Configuration."
         )
         if filename:
+            if not filename.endswith(".cfg"):
+                filename += ".cfg"
             self._mmc.saveSystemConfiguration(filename)
+
+            # saveSystemConfiguration does not save the pixel size config so here
+            # we add to the saved .cfg filesave also any pixel size config.
+            if px_configs := self._mmc.getAvailablePixelSizeConfigs():
+                with open(filename, "a") as f:
+                    f.write("# PixelSize settings")
+                    for px_config in px_configs:
+                        data = self._mmc.getPixelSizeConfigData(px_config)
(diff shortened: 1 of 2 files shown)


tlambert03:
made a [small edit](https://github.com/pymmcore-plus/pymmcore-widgets/pull/150/commits/be8a4f2300a63b6ecf2dd0663c190df791a59a99) here to use pytest tmp_path fixture rather than creating our own temporary directory

src/pymmcore_widgets/_group_preset_widget/_group_preset_table_widget.py
```diff
@@ -401,8 +401,27 @@ def _save_cfg(self) -> None:
             self, "Save Micro-Manager Configuration."
         )
         if filename:
+            if not filename.endswith(".cfg"):
+                filename += ".cfg"
             self._mmc.saveSystemConfiguration(filename)
 
+            # saveSystemConfiguration does not save the pixel size config so here
+            # we add to the saved .cfg filesave also any pixel size config.
+            if px_configs := self._mmc.getAvailablePixelSizeConfigs():
```
tlambert03, line 410: it looks like this would append  a new `# PixelSize settings` section every time the file is saved. (so one might end up with a ton of those sections over time... can you confirm that?

fdrgsp: That's right, I didn't think about that. I think we should overwrite it...

fdrgsp: @tlambert03 Actually, since we create a new cfg _without_ any px info every time we call `_save_cfg`, there will always be a single `# PixelSize settings`.

tlambert03: ok sounds good

fdrgsp:
moving to `pymmcore_plus`.

PR #154, opened 2023-07-25
fix: reduce widget freezing

wl-stepp:
Hello pymmcore-widgets team,

I had the MDA widget freeze on me pretty regularly. For me it seems to be much more stable with these changes. Maybe worth thinking about.

Here is my pip list in case you see a problem there:
```
appdirs                       1.4.4
click                         8.1.6
colorama                      0.4.6
fonticon-materialdesignicons6 6.9.96
loguru                        0.7.0
markdown-it-py                3.0.0
mdurl                         0.1.2
mypy-extensions               1.0.0
numpy                         1.25.1
packaging                     23.1
Pint                          0.22
pip                           22.3
psygnal                       0.9.1
pydantic                      1.10.12
Pygments                      2.15.1
pymmcore                      10.3.0.71.0
pymmcore-plus                 0.6.7
pymmcore-widgets              0.3.1.dev21+g90706e6
PyQt6                         6.5.2
PyQt6-Qt6                     6.5.2
PyQt6-sip                     13.5.2
QtPy                          2.3.1
rich                          13.4.2
setuptools                    65.5.0
superqt                       0.4.1
typer                         0.9.0
typing_extensions             4.7.1
useq-schema                   0.3.0
win32-setctime                1.1.0
wrapt                         1.15.0
```


tlambert03 (approved):
this looks great to me, thanks for the PR.

Tests are failing because of the way `test_mda_widgets.py::test_mda_methods` was designed... can you just make the following change to the test right here:

https://github.com/pymmcore-plus/pymmcore-widgets/blob/90706e6734cf57472bd67f950da02fadaffc450f/tests/test_mda_widget.py#L114

```diff
- wdg._on_mda_started()
+ seq = MDASequence()
+ global_mmcore.mda.events.sequenceStarted.emit(seq)
```

and then as well below 

```diff
- wdg._on_mda_finished()
+ global_mmcore.mda.events.sequenceFinished.emit(seq)
```

and then go ahead and just delete the lines you commented out

fdrgsp:
thanks @wl-stepp! @tlambert03 I think that if we do this change we probably have to remove the `MDAWidget` `include_run_button` argument or move it to `_MDAControlButtons` if we want to maintain the functionality. What do you think?

tlambert03:
that's a good point @fdrgsp ... I think the _main_ thing that we wanted to be user-configurable was more the connection of the button to `self._mmc.run_mda()` (and less the visibility of the button).  So, perhaps just changing the name of the parameter to `connect_run_button` would be enough, and we keep the rest.

I'm also fine just removing the parameter altogether.  The point was just to make it possible for someone to use this widget without being obligated to using the full pymmcore_plus mda runner

fdrgsp:
> just changing the name of the parameter to connect_run_button would be enough, and we keep the rest

The issue is that if we don't control the show/hide state of the run button with the `include_run_button` parameter and we only hide it at the beginning, since the run button will be connected to the `sequenceStarted` and `sequenceFinished` it will be shown after the sequence is finished.

> The point was just to make it possible for someone to use this widget without being obligated to using the full pymmcore_plus mda runner

I think because of this is good to keep this functionality. Maybe we can pass a `bool` `connect_to_core` parameter to `_MDAControlButtons` and connect the signals only if the bool is true...

```py
class MDAWidget(QWidget):
    def __init__(
        self, *, parent: QWidget | None = None, connect_to_core: bool = False, mmcore: CMMCorePlus | None = None,
    ) -> None:
    self. connect_to_core = connect_to_core
    ...
    self.buttons_wdg = _MDAControlButtons(connect_to_core=self._connect_run_button)
```

```py
class _MDAControlButtons(QWidget):
    def __init__(self, *, parent: QWidget | None = None, connect_to_core: bool = False) -> None:
        super().__init__(parent=parent)
        ...
        if connect_to_core:
            self._mmc.mda.events.sequenceStarted.connect(self._on_mda_started)
            self._mmc.mda.events.sequenceFinished.connect(self._on_mda_finished)
```

src/pymmcore_widgets/_mda/_general_mda_widgets.py
```diff
@@ -83,10 +75,40 @@ def _create_btns_gui(self) -> None:
...
+    def _on_mda_started(self) -> None:
+        self.run_button.hide()
+        self.pause_button.show()
+        self.cancel_button.show()
+
+    def _on_mda_finished(self) -> None:
+        self.run_button.show()
+        self.pause_button.hide()
+        self.cancel_button.hide()
+        self.pause_button.setText("Pause")
```
wl-stepp, line 87: If you canceled out of a pause, the button stayed at "Resume" for the next run

src/pymmcore_widgets/_mda/_mda_widget.py
```diff
@@ -205,8 +206,13 @@ def __init__(
         self._mmc.events.channelGroupChanged.connect(self._on_channel_group_changed)
         # connect run button
         if self._include_run_button:
+            self.buttons_wdg = _MDAControlButtons()
             self.buttons_wdg.run_button.clicked.connect(self._on_run_clicked)
             self.buttons_wdg.run_button.show()
+            # connect buttons
+            self.buttons_wdg.pause_button.released.connect(self._mmc.mda.toggle_pause)
+            self.buttons_wdg.cancel_button.released.connect(self._mmc.mda.cancel)
+            acq_order_run_layout.addWidget(self.buttons_wdg)
```
wl-stepp, line 215: Don't even make the widget if it's not going to be used.

fdrgsp:
> Let me know what you think.

@wl-stepp I do like it!

src/pymmcore_widgets/_mda/_mda_widget.py
```diff
@@ -21,6 +22,7 @@
 from ._channel_table_widget import ChannelTable
 from ._checkable_tabwidget_widget import CheckableTabWidget
 from ._general_mda_widgets import (
+    AcquisitionOrderWidget,
```
fdrgsp, line 25:
```suggestion
   _AcquisitionOrderWidget,
```

src/pymmcore_widgets/_mda/_mda_widget.py
```diff
@@ -141,16 +143,17 @@ def __init__(
...
-        self.buttons_wdg.cancel_button.hide()
-        self.buttons_wdg.run_button.hide()
 
         # savle load widget
         self._save_load = SaveLoadSequenceWidget()
         self._save_load._save_button.clicked.connect(self._save_sequence)
         self._save_load._load_button.clicked.connect(self._load_sequence)
 
+        # Acquisition order widget
+        self.acquisition_order_widget = AcquisitionOrderWidget()
```
fdrgsp, line 153:
```suggestion
        self.acquisition_order_widget = _AcquisitionOrderWidget()
```

fdrgsp:
Thanks @wl-stepp!

PR #157, opened 2023-07-26
perf: Improve speed of mda widget

ianhi:
This brings significant speed ups (e.g. 1 min -> 7 seconds) to the mda widget and some of it's subcomponents. The most significant thing I was running into as a user was that we were having the UI freeze as we tried to type out larger numbers for timepoints. This even resulted in us having to close everything and restart at least once. This was caused by slowness in the min total time calculation, and it being called over and over as we typed into the spinbox.

### DIFF
--- a/src/pymmcore_widgets/_mda/_mda_widget.py
+++ b/src/pymmcore_widgets/_mda/_mda_widget.py
@@ -308,6 +308,9 @@ class MDAWidget(QWidget):
             MDASequence state in the form of a dict, MDASequence object, or a str or
             Path pointing to a sequence.yaml file
         """
+        # TODO: prevent _update_total_time from being called until
+        # all subcomponents have been set
+
         # sourcery skip: low-code-quality
         if isinstance(state, (str, Path)):
             state = MDASequence.parse_file(state)
--- a/src/pymmcore_widgets/_mda/_mda_widget.py
+++ b/src/pymmcore_widgets/_mda/_mda_widget.py
@@ -486,13 +486,14 @@ class MDAWidget(QWidget):
         total_time: float = 0.0
         _per_timepoints: dict[int, float] = {}
         t_per_tp_msg = ""
+        _uses_time = self._uses_time()
 
         for e in self.get_state():
             if e.exposure is None:
                 continue
 
             total_time = total_time + (e.exposure / 1000)
-            if self._uses_time():
+            if _uses_time:
                 _t = e.index["t"]
                 _exp = e.exposure / 1000
                 _per_timepoints[_t] = _per_timepoints.get(_t, 0) + _exp
--- a/src/pymmcore_widgets/_mda/_mda_widget.py
+++ b/src/pymmcore_widgets/_mda/_mda_widget.py
@@ -465,7 +465,7 @@ class MDAWidget(QWidget):
 
     def _uses_time(self) -> bool:
         """Hacky method to check whether the timebox is selected with any timepoints."""
-        has_phases = self.time_widget.value()["phases"]  # type: ignore
+        has_phases = self.time_widget._table.rowCount()
         return bool(self.t_cbox.isChecked() and has_phases)
 
     def _update_total_time(self) -> None:
--- a/src/pymmcore_widgets/_mda/_positions_table_widget.py
+++ b/src/pymmcore_widgets/_mda/_positions_table_widget.py
@@ -277,6 +277,7 @@ class PositionTable(QWidget):
         spin.setButtonSymbols(QAbstractSpinBox.ButtonSymbols.NoButtons)
         spin.setValue(value)
         spin.wheelEvent = lambda event: None  # block mouse scroll
+        spin.setKeyboardTracking(False)
         self._table.setCellWidget(row, col, spin)
 
     def _add_grid_buttons(self, row: int | None, col: int | None) -> None:
--- a/src/pymmcore_widgets/_mda/_positions_table_widget.py
+++ b/src/pymmcore_widgets/_mda/_positions_table_widget.py
@@ -538,35 +538,35 @@ class PositionTable(QWidget):
             By default True. If True, the current positions list is cleared before the
             specified one is added.
         """
-        if clear:
-            self.clear()
-
         if not isinstance(positions, Sequence):
             raise TypeError("The 'positions' arguments has to be a 'Sequence'.")
 
         if not self._mmc.getXYStageDevice() and not self._mmc.getFocusDevice():
             raise ValueError("No XY and Z Stage devices loaded.")
 
-        for position in positions:
-            if isinstance(position, Position):
-                position = cast("PositionDict", position.dict())
+        with signals_blocked(self):
+            if clear:
+                self.clear()
+            for position in positions:
+                if isinstance(position, Position):
+                    position = cast("PositionDict", position.dict())
 
-            if not isinstance(position, dict):
-                continue
+                if not isinstance(position, dict):
+                    continue
 
-            name = position.get("name")
-            x, y, z = (position.get("x"), position.get("y"), position.get("z"))
-            self._add_table_row(name or f"{POS}000", x, y, z)
-            if pos_seq := position.get("sequence"):
-                self._advanced_cbox.setChecked(True)
-                if isinstance(pos_seq, MDASequence):
-                    grid_plan = pos_seq.grid_plan.dict()
-                else:
-                    grid_plan = pos_seq.get("grid_plan")
-                if grid_plan:
-                    self._add_grid_plan(grid_plan, self._table.rowCount() - 1)
+                name = position.get("name")
+                x, y, z = (position.get("x"), position.get("y"), position.get("z"))
+                self._add_table_row(name or f"{POS}000", x, y, z)
+                if pos_seq := position.get("sequence"):
+                    self._advanced_cbox.setChecked(True)
+                    if isinstance(pos_seq, MDASequence):
+                        grid_plan = pos_seq.grid_plan.dict()
+                    else:
+                        grid_plan = pos_seq.get("grid_plan")
+                    if grid_plan:
+                        self._add_grid_plan(grid_plan, self._table.rowCount() - 1)
 
-            self.valueChanged.emit()
+        self.valueChanged.emit()
 
     def _save_positions(self) -> None:
         if not self._table.rowCount() or not self.value():
--- a/src/pymmcore_widgets/_mda/_time_plan_widget.py
+++ b/src/pymmcore_widgets/_mda/_time_plan_widget.py
@@ -271,9 +271,11 @@ class TimePlanWidget(QWidget):
         with signals_blocked(self):
             for phase in phases:
                 if not isinstance(phase, TIntervalLoops):
-                    raise ValueError("Time dicts must have both 'interval' and 'loops'.")
+                    raise ValueError(
+                        "Time dicts must have both 'interval' and 'loops'."
+                    )
                 self._create_new_row(interval=phase.interval, loops=phase.loops)
-        self.valueChanged()
+        self.valueChanged.emit()
 
     def _disconnect(self) -> None:
         self._mmc.events.systemConfigurationLoaded.disconnect(self._clear)
--- a/src/pymmcore_widgets/_mda/_time_plan_widget.py
+++ b/src/pymmcore_widgets/_mda/_time_plan_widget.py
@@ -271,13 +271,12 @@ class TimePlanWidget(QWidget):
         with signals_blocked(self):
             self._clear()
             phases = tp.phases if isinstance(tp, MultiPhaseTimePlan) else [tp]
-            with signals_blocked(self):
-                for phase in phases:
-                    if not isinstance(phase, TIntervalLoops):
-                        raise ValueError(
-                            "Time dicts must have both 'interval' and 'loops'."
-                        )
-                    self._create_new_row(interval=phase.interval, loops=phase.loops)
+            for phase in phases:
+                if not isinstance(phase, TIntervalLoops):
+                    raise ValueError(
+                        "Time dicts must have both 'interval' and 'loops'."
+                    )
+                self._create_new_row(interval=phase.interval, loops=phase.loops)
         self.valueChanged.emit()
 
     def _disconnect(self) -> None:
--- a/src/pymmcore_widgets/_mda/_time_plan_widget.py
+++ b/src/pymmcore_widgets/_mda/_time_plan_widget.py
@@ -23,6 +23,7 @@ from qtpy.QtWidgets import (
     QWidget,
 )
 from superqt import QQuantity, fonticon
+from superqt.utils import signals_blocked
 from useq import MDASequence, MultiPhaseTimePlan, TIntervalLoops
 
 if TYPE_CHECKING:
@@ -267,10 +268,12 @@ class TimePlanWidget(QWidget):
         if tp is None:
             return
         phases = tp.phases if isinstance(tp, MultiPhaseTimePlan) else [tp]
-        for phase in phases:
-            if not isinstance(phase, TIntervalLoops):
-                raise ValueError("Time dicts must have both 'interval' and 'loops'.")
-            self._create_new_row(interval=phase.interval, loops=phase.loops)
+        with signals_blocked(self):
+            for phase in phases:
+                if not isinstance(phase, TIntervalLoops):
+                    raise ValueError("Time dicts must have both 'interval' and 'loops'.")
+                self._create_new_row(interval=phase.interval, loops=phase.loops)
+        self.valueChanged()
 
     def _disconnect(self) -> None:
         self._mmc.events.systemConfigurationLoaded.disconnect(self._clear)
--- a/src/pymmcore_widgets/_mda/_time_plan_widget.py
+++ b/src/pymmcore_widgets/_mda/_time_plan_widget.py
@@ -181,6 +181,7 @@ class TimePlanWidget(QWidget):
         time_spin.setValue(loops or 1)
         time_spin.setButtonSymbols(QAbstractSpinBox.ButtonSymbols.NoButtons)
         time_spin.setAlignment(Qt.AlignmentFlag.AlignCenter)
+        time_spin.setKeyboardTracking(False)
         time_spin.valueChanged.connect(self.valueChanged)
 
         idx = self._table.rowCount()
--- a/src/pymmcore_widgets/_mda/_time_plan_widget.py
+++ b/src/pymmcore_widgets/_mda/_time_plan_widget.py
@@ -263,19 +263,20 @@ class TimePlanWidget(QWidget):
             ) `interval` key is not a `timedelta` object, it will be converted to a
             timedelta object and will be considered as expressed in seconds.
         """
-        self._clear()
-
         tp = MDASequence(time_plan=t_plan).time_plan
         if tp is None:
+            # should we raise/warn here?
             return
-        phases = tp.phases if isinstance(tp, MultiPhaseTimePlan) else [tp]
         with signals_blocked(self):
-            for phase in phases:
-                if not isinstance(phase, TIntervalLoops):
-                    raise ValueError(
-                        "Time dicts must have both 'interval' and 'loops'."
-                    )
-                self._create_new_row(interval=phase.interval, loops=phase.loops)
+            self._clear()
+            phases = tp.phases if isinstance(tp, MultiPhaseTimePlan) else [tp]
+            with signals_blocked(self):
+                for phase in phases:
+                    if not isinstance(phase, TIntervalLoops):
+                        raise ValueError(
+                            "Time dicts must have both 'interval' and 'loops'."
+                        )
+                    self._create_new_row(interval=phase.interval, loops=phase.loops)
         self.valueChanged.emit()
 
     def _disconnect(self) -> None:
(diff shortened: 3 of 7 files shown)


src/pymmcore_widgets/_mda/_mda_widget.py
```diff
@@ -308,6 +308,9 @@ def set_state(self, state: dict | MDASequence | str | Path) -> None:
             MDASequence state in the form of a dict, MDASequence object, or a str or
             Path pointing to a sequence.yaml file
         """
+        # TODO: prevent _update_total_time from being called until
+        # all subcomponents have been set
```
ianhi, line 312: @tlambert03 any suggestions here? I looked at https://pyapp-kit.github.io/superqt/utilities/throttling/#superqt.utils._throttler.GenericSignalThrottler.throttle but that didn't seem right as the signals are already connected to the existing function. We also can't block them in case something else is listening, it seems like we need to block a specific function from receiving the signals

src/pymmcore_widgets/_mda/_mda_widget.py
```diff
@@ -486,13 +489,14 @@ def _update_total_time(self) -> None:
...
         t_per_tp_msg = ""
+        _uses_time = self._uses_time()
 
         for e in self.get_state():
             if e.exposure is None:
                 continue
 
             total_time = total_time + (e.exposure / 1000)
-            if self._uses_time():
+            if _uses_time:
```
ianhi, line 499: I line profiled this and this was taking ~75% of the time in this function.

tlambert03: wow.  good find

src/pymmcore_widgets/_mda/_positions_table_widget.py
```diff
@@ -538,35 +538,35 @@ def set_state(
...
         if not self._mmc.getXYStageDevice() and not self._mmc.getFocusDevice():
             raise ValueError("No XY and Z Stage devices loaded.")
 
-        for position in positions:
-            if isinstance(position, Position):
-                position = cast("PositionDict", position.dict())
-
-            if not isinstance(position, dict):
-                continue
+        with signals_blocked(self):
```
ianhi, line 547: delaying the valueChanged signal till the very end is effectively the maximum speed up possible here.

tlambert03:
fantastic.  is there a test you could add (maybe to assert that no more than one signal is emitted?)  If it's cumbersome, no biggie, but if it's easy great

ianhi:
hmm, trying this out it's now much more useable for the timelapses I'm doing (500 time points) but still freezes the ui for a few seconds, and becomes unusable when you get into the thousands range. Two potential improvements:

1. Throttle the min_time func
2. have the time calculation run in a separate worker thread

I suppose thre's also a deeper fix of figuring out a way to estimate total time without iterating over every event.



>  is there a test you could add (maybe to assert that no more than one signal is emitted?)

I'm not exactly sure what you want tested. I think i'm misunderstanding but that feels as though it would result in a test of the functionality of `signals_blocked`

tlambert03:
> I'm not exactly sure what you want tested. I think i'm misunderstanding but that feels as though it would result in a test of the functionality of signals_blocked

sure, we know signals blocks works, but until this PR, we weren't using it.  So, to prevent regressions in the future, if you've found a specific signal or set thereof that is a proxy for bad code practices, a test would ensure that we didn't slip into excess signal emission in the future.  but like I said, no biggie if it feels contrived.

> trying this out it's now much more useable for the timelapses I'm doing (500 time points) but still freezes the ui for a few seconds, and becomes unusable when you get into the thousands range

can you tell me exactly what I need to do to see this?  Is it simply in interacting with the widget and typing into one of the fields?

Is this all just because of the nice user feedback to alert them to possible warnings?  I feel like that part of the widget was always a "nice feature" but could be reconsidered altogether if it's actually causing performance issues.

src/pymmcore_widgets/_mda/_mda_widget.py
```diff
@@ -486,13 +489,14 @@ def _update_total_time(self) -> None:
         total_time: float = 0.0
         _per_timepoints: dict[int, float] = {}
         t_per_tp_msg = ""
+        _uses_time = self._uses_time()
 
         for e in self.get_state():
```
fdrgsp, line 494: @ianhi @tlambert03 I think the main issue is here when we call `self.get_state()`. I basically already tried to fix it in this branch (https://github.com/fdrgsp/pymmcore-widgets/tree/update_time_in_thread) using a different thread and it works great on the microscope (no freeze at all). However it gives me the `_run_after_each_test` error in test so I did not open a PR yet.

ianhi: Speeding up `get_state` would be the next big win, and also generically a good thing. However, as it stands the uses_time was actually the most costly part. You can see this from the `line_profiler` breakdown:

```
Timer unit: 1e-09 s

Total time: 18.2211 s
File: /home/ian/Documents/oss/micro/pymmcore-widgets/src/pymmcore_widgets/_mda/_mda_widget.py
Function: _update_total_time at line 468

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
   468                                               def _update_total_time(self) -> None:
   469                                                   """Calculate the minimum total acquisition time info."""
   470                                                   # TODO: fix me!!!!!
   471         1      70016.0  70016.0      0.0          if self._mmc.getChannelGroup() and self._mmc.getCurrentConfig(
   472         1       4312.0   4312.0      0.0              self._mmc.getChannelGroup()
   473                                                   ):
   474         1      13291.0  13291.0      0.0              if self.ch_cbox.isChecked() and not self.channel_widget._table.rowCount():
   475                                                           self.time_lbl._total_time_lbl.setText(
   476                                                               "Minimum total acquisition time: 0 sec."
   477                                                           )
   478                                                           return
   479                                           
   480                                                   elif not self.ch_cbox.isChecked() or not self.channel_widget._table.rowCount():
   481                                                       self.time_lbl._total_time_lbl.setText(
   482                                                           "Minimum total acquisition time: 0 sec."
   483                                                       )
   484                                                       return
   485                                           
   486         1        689.0    689.0      0.0          total_time: float = 0.0
   487         1        480.0    480.0      0.0          _per_timepoints: dict[int, float] = {}
   488         1        336.0    336.0      0.0          t_per_tp_msg = ""
   489                                                   # _uses_time = self._uses_time()
   490                                           
   491     36000 4822693525.0 133963.7     26.5          for e in self.get_state():
   492     36000   21320506.0    592.2      0.1              if e.exposure is None:
   493                                                           continue
   494                                           
   495     36000   28594420.0    794.3      0.2              total_time = total_time + (e.exposure / 1000)
   496     36000 13228951745.0 367470.9     72.6              if self._uses_time():
   497     36000   54440471.0   1512.2      0.3                  _t = e.index["t"]
   498     36000   22185299.0    616.3      0.1                  _exp = e.exposure / 1000
   499     36000   42204671.0   1172.4      0.2                  _per_timepoints[_t] = _per_timepoints.get(_t, 0) + _exp
   500                                           
   501         1        443.0    443.0      0.0          if _per_timepoints:
   502         1     320968.0 320968.0      0.0              time_value = self.time_widget.value()
   503                                           
   504         1        263.0    263.0      0.0              intervals = []
   505         1        597.0    597.0      0.0              for phase in time_value["phases"]:  # type: ignore
   506         1       1539.0   1539.0      0.0                  interval = phase["interval"].total_seconds()
   507         1        432.0    432.0      0.0                  intervals.append(interval)
   508         1        452.0    452.0      0.0                  if phase.get("loops") is not None:
   509         1        970.0    970.0      0.0                      total_time = total_time + (phase["loops"] - 1) * interval
   510                                                           else:
   511                                                               total_time = total_time + phase["duration"].total_seconds()
   512                                           
   513                                                       # check if the interval(s) is smaller than the sum of the exposure times
   514         1       7124.0   7124.0      0.0              sum_ch_exp = sum(
   515                                                           (c["exposure"] / 1000)
   516         1      82319.0  82319.0      0.0                  for c in self.channel_widget.value()
   517                                                           if c["exposure"] is not None
   518                                                       )
   519         1        322.0    322.0      0.0              for i in intervals:
   520         1        738.0    738.0      0.0                  if 0 < i < sum_ch_exp:
   521                                                               self.time_widget.setWarningVisible(True)
   522                                                               break
   523                                                           else:
   524         1       5945.0   5945.0      0.0                      self.time_widget.setWarningVisible(False)
   525                                           
   526                                                       # group by time
   527         1     106683.0 106683.0      0.0              _group_by_time: dict[float, list[int]] = {
   528                                                           n: [k for k in _per_timepoints if _per_timepoints[k] == n]
   529         1      21566.0  21566.0      0.0                  for n in set(_per_timepoints.values())
   530                                                       }
   531                                           
   532         1        282.0    282.0      0.0              t_per_tp_msg = "Minimum acquisition time per timepoint: "
   533                                           
   534         1        884.0    884.0      0.0              if len(_group_by_time) == 1:
   535         1        428.0    428.0      0.0                  t_per_tp_msg = (
   536         1        545.0    545.0      0.0                      f"\n{t_per_tp_msg}"
   537         1      35895.0  35895.0      0.0                      f"{fmt_timedelta(timedelta(seconds=_per_timepoints[0]))}"
   538                                                           )
   539                                                       else:
   540                                                           acq_min = timedelta(seconds=min(_per_timepoints.values()))
   541                                                           t_per_tp_msg = (
   542                                                               f"\n{t_per_tp_msg}{fmt_timedelta(acq_min)}"
   543                                                               if self._uses_time()
   544                                                               else ""
   545                                                           )
   546                                                   else:
   547                                                       t_per_tp_msg = ""
   548                                                       self.time_widget.setWarningVisible(False)
   549                                           
   550         1        169.0    169.0      0.0          _min_tot_time = (
   551         1        336.0    336.0      0.0              "Minimum total acquisition time: "
   552         1      12258.0  12258.0      0.0              f"{fmt_timedelta(timedelta(seconds=total_time))}"
   553                                                   )
   554         1       8285.0   8285.0      0.0          self.time_lbl._total_time_lbl.setText(f"{_min_tot_time}{t_per_tp_msg}")

```

ianhi:
> So, to prevent regressions in the future, if you've found a specific signal or set thereof that is a proxy for bad code practices, a test would ensure that we didn't slip into excess signal emission in the future. but like I said, no biggie if it feels contrived.

Gotcha, ya I think I can add something like that for the `set_state` methods. Can probably catch them all via the mda_widget's `set_state` method even.

> can you tell me exactly what I need to do to see this? Is it simply in interacting with the widget and typing into one of the fields?

run the `examples/mda_widget.py` example, then add a few channels, a z plan, some positions then start typing out larger and large numbers into a time plan. e.g. `5` -> `50`, `500` -> `5000` and it will have progressively longer freezes.


> Is this all just because of the nice user feedback to alert them to possible warnings? I feel like that part of the widget was always a "nice feature" but could be reconsidered altogether if it's actually causing performance issues.


I'm not even sure it's about warnings - I think the user feedback at fault is the just the calculation of this:
image.

Some of the improvements here (i.e. all of the `set_state` ) should be kept regardless, but if we can't fully solve the performance here then dropping that time estimate would be good. That said I think we could the take the performance improvements here and couple them with federico's threaded version and leave it in. Although it should be clear when the value is stale - i.e. go blank until it's finished calculating.

src/pymmcore_widgets/_mda/_time_plan_widget.py
```diff
@@ -180,6 +181,7 @@ def _create_new_row(
         time_spin.setValue(loops or 1)
         time_spin.setButtonSymbols(QAbstractSpinBox.ButtonSymbols.NoButtons)
         time_spin.setAlignment(Qt.AlignmentFlag.AlignCenter)
+        time_spin.setKeyboardTracking(False)
```
ianhi, line 184: could probably also stand to bring a similar thing to all the other subcomponents like here:
https://github.com/pymmcore-plus/pymmcore-widgets/blob/58f35f4c3fdcdf8e830c301d4a3de8215c1a948b/src/pymmcore_widgets/_mda/_channel_table_widget.py#L263-L264

src/pymmcore_widgets/_mda/_mda_widget.py
tlambert03: > I think the main issue is here when we call self.get_state(). I basically already tried to fix it in this branch

I don't think the issue is *calling* `get_state()`, I think the issue is iterating it.  It's an easy trap to fall into, but basically we should pretty much try to avoid ever resorting to iterating over an MDASequence.  Only the final experiment should do that

ianhi:
> can you tell me exactly what I need to do to see this? 

You can speed up the manual iteration process a bit with the below script. Though beware that without the improvements here to `set_state` the `set_state` call will take quite awhile (I discovered those issues by trying to debug the min_time issue).

<details>
 <summary><b>Self contained script </b></summary>

```python
from qtpy.QtWidgets import (
    QApplication,
    QGroupBox,
    QHBoxLayout,
    QLabel,
    QVBoxLayout,
    QWidget,
)
from useq import MDAEvent, MDASequence

from pymmcore_widgets import MDAWidget
import datetime

from useq import (
    Channel,
    MDASequence,
    MultiPhaseTimePlan,
    Position,
    TIntervalLoops,
    ZRangeAround,
)



class MDA(QWidget):
    """An example of using the MDAWidget to create and acquire a useq.MDASequence.

    The `MDAWidget` provides a GUI to construct a `useq.MDASequence` object.
    This object describes a full multi-dimensional acquisition;

    In this example, we set the `MDAWidget` parameter `include_run_button` to `True`,
    meaning that a `run` button is added to the GUI. When pressed, a `useq.MDASequence`
    is first built depending on the GUI values and is then passed to the
    `CMMCorePlus.run_mda` to actually execute the acquisition.

    For details of the corresponding schema and methods, see
    https://github.com/pymmcore-plus/useq-schema and
    https://github.com/pymmcore-plus/pymmcore-plus.

    In this example, we've also connected callbacks to the CMMCorePlus object's `mda`
    events to print out the current state of the acquisition.
    """

    def __init__(self) -> None:
        super().__init__()

        # get the CMMCore instance and load the default config
        self.mmc = CMMCorePlus.instance()
        self.mmc.loadSystemConfiguration()

        # connect MDA acquisition events to local callbacks
        # in this example we're just printing the current state of the acquisition
        self.mmc.mda.events.sequenceStarted.connect(self._on_start)
        self.mmc.mda.events.frameReady.connect(self._on_frame)
        self.mmc.mda.events.sequenceFinished.connect(self._on_end)
        self.mmc.mda.events.sequencePauseToggled.connect(self._on_pause)

        # instantiate the MDAWidget, and a couple lables for feedback
        self.mda = MDAWidget(include_run_button=True)
        self.current_sequence = QLabel('... enter info and click "Run"')
        self.current_event = QLabel("... current event info will appear here")

        # below here is just GUI layout stuff
        mda_wdg = QGroupBox()
        mda_wdg.setMaximumWidth(600)
        mda_wdg.setLayout(QVBoxLayout())
        mda_wdg.layout().setContentsMargins(0, 0, 0, 0)
        mda_wdg.layout().addWidget(self.mda)

        lbl_wdg = QGroupBox()
        lbl_wdg.setMinimumWidth(275)
        lbl_wdg.setLayout(QVBoxLayout())
        lbl_wdg.layout().addWidget(QLabel(text="<h3>ACQUISITION SEQUENCE</h3>"))
        lbl_wdg.layout().addWidget(self.current_sequence)
        lbl_wdg.layout().addWidget(QLabel(text="<h3>ACQUISITION EVENT</h3>"))
        lbl_wdg.layout().addWidget(self.current_event)

        self.setLayout(QHBoxLayout())
        self.layout().addWidget(mda_wdg)
        self.layout().addWidget(lbl_wdg)
        self.resize(900, 800)

    def _on_start(self, sequence: MDASequence) -> None:
        """Called when the MDA sequence starts."""
        self.current_sequence.setText(sequence.yaml())

    def _on_frame(self, image, event: MDAEvent) -> None:
        """Called each time a frame is acquired."""
        self.current_event.setText(
            f"index: {event.index}\n"
            f"channel: {event.channel.config}\n"
            f"exposure: {event.exposure}\n"
            f"pos_name: {event.pos_name}\n"
            f"xyz: ({event.x_pos}, {event.y_pos}, {event.z_pos})\n"
        )

    def _on_end(self) -> None:
        """Called when the MDA sequence ends."""
        self.current_event.setText("Finished!")

    def _on_pause(self, state: bool) -> None:
        """Called when the MDA is paused."""
        txt = "Paused..." if state else "Resumed!"
        self.current_event.setText(txt)

seq = MDASequence(
    channels=(
        Channel(
            config="Cy5",
            group="Channel",
            exposure=10.0,
            do_stack=True,
            z_offset=0.0,
            acquire_every=1,
            camera=None,
        ),
        Channel(
            config="DAPI",
            group="Channel",
            exposure=10.0,
            do_stack=True,
            z_offset=0.0,
            acquire_every=1,
            camera=None,
        ),
        Channel(
            config="FITC",
            group="Channel",
            exposure=10.0,
            do_stack=True,
            z_offset=0.0,
            acquire_every=1,
            camera=None,
        ),
    ),
    stage_positions=(
        Position(x=0.0, y=0.0, z=0.0, name="Pos000", sequence=None),
        Position(x=0.0, y=0.0, z=0.0, name="Pos001", sequence=None),
        Position(x=0.0, y=0.0, z=0.0, name="Pos002", sequence=None),
        Position(x=0.0, y=0.0, z=0.0, name="Pos003", sequence=None),
    ),
    time_plan=MultiPhaseTimePlan(
        prioritize_duration=False,
        phases=[
            TIntervalLoops(
                prioritize_duration=False,
                interval=datetime.timedelta(seconds=1),
                loops=500, # CHANGE THIS LINE - to modify how long things take
            )
        ],
    ),
    z_plan=ZRangeAround(go_up=True, range=5.0, step=1.0),
)

app = QApplication([])
frame = MDA()
frame.mda.set_state(seq)
frame.show()
app.exec_()
```
</details>

src/pymmcore_widgets/_mda/_mda_widget.py
tlambert03: also, from looking at it, uses_time could probably be sped up slightly by changing it from this:

```py
    def _uses_time(self) -> bool:
        """Hacky method to check whether the timebox is selected with any timepoints."""
        has_phases = self.time_widget.value()["phases"]  # type: ignore
        return bool(self.t_cbox.isChecked() and has_phases)
```

to this

```py
    def _uses_time(self) -> bool:
        """Hacky method to check whether the timebox is selected with any timepoints."""
        has_phases = self.time_widget._table.rowCount()
        return bool(self.t_cbox.isChecked() and has_phases)
```

tlambert03: also, for what it's worth... i think the whole time estimation thing might be better off as an `estimate_time()` method on MDASequence in useq.

ianhi: The new `_uses_time` is a good microoptimization. That function gets a 205x speed increase from 122 microseconds -> 595 nanoseconds. Sadly we only call it once now so not as much of a speed up as we might've gotten before :(

ianhi:
Added a new commit setting `keyboardTracking` to be False everywhere in the mda. So long as there are expensive calculations triggered by changes here we need this. If the downstream effects get radically sped up then we could remove these.

tlambert03:
let me know when you're done with this and want a final review/merge

ianhi:
ok I'm done here. This is ready for final review/merge.

If we decide it's necessary we can always add the threading in a separate PR

src/pymmcore_widgets/_mda/_mda_widget.py
tlambert03: yeah, a debouncer would be a good idea here.  They would need to be debounced at connection time though, so it would look something like 

```py
from superqt.utils import qdebounced

self.channel_widget.valueChanged.connect(qdebounced(self._update_total_time))
```

but of course, that's hooked up a million times... so it would need to go everywhere.  can follow up in a later PR

src/pymmcore_widgets/_mda/_time_plan_widget.py
```diff
@@ -261,16 +264,21 @@ def set_state(self, t_plan: dict) -> None:
...
             return
-        phases = tp.phases if isinstance(tp, MultiPhaseTimePlan) else [tp]
-        for phase in phases:
-            if not isinstance(phase, TIntervalLoops):
-                raise ValueError("Time dicts must have both 'interval' and 'loops'.")
-            self._create_new_row(interval=phase.interval, loops=phase.loops)
+        with signals_blocked(self):
+            self._clear()
+            phases = tp.phases if isinstance(tp, MultiPhaseTimePlan) else [tp]
+            with signals_blocked(self):
```
tlambert03, line 274: is this duplicate usage of `signals_blocked` intentional @ianhi?

tlambert03 (approved):
looks good to me, will merge after hearing back about the seemingly duplicated signal blocker

src/pymmcore_widgets/_mda/_time_plan_widget.py
tlambert03: i deleted and tests still pass... so will go with it.  Let me know if i'm mistaken

ianhi: this was correct to remove. I must've missed the internal one when i came back to include the `clear` inside of the blocker

tlambert03: thanks for confirming!

Issue #158, opened 2023-07-27
test: add test of upstream napari-micromanager


ianhi:
Since napari-micromanager is dependent on everything here I'm finding that changes here are breaking things there. For example this line:

https://github.com/pymmcore-plus/pymmcore-widgets/blob/58f35f4c3fdcdf8e830c301d4a3de8215c1a948b/src/pymmcore_widgets/_mda/_mda_widget.py#L219

which was recently added in #140 results in this error for napari-micromanager:

```
AttributeError                            Traceback (most recent call last)
File ~/Documents/oss/micro/napari-micromanager/src/napari_micromanager/_gui_objects/_toolbar.py:369, in MicroManagerToolbar._show_dock_widget(self=<napari_micromanager.main_window.MainWindow(0x55f386059e80)>, key='MDA')
    363 except KeyError as e:
    364     raise KeyError(
    365         "Not a recognized dock widget key. "
    366         f"Must be one of {list(DOCK_WIDGETS)} "
    367         " or the `whatsThis` property of a `sender` `QPushButton`."
    368     ) from e
--> 369 wdg = wdg_cls(parent=self, mmcore=self._mmc)
        self = <napari_micromanager.main_window.MainWindow(0x55f386059e80) at 0x7f538811a7c0>
        wdg_cls = <class 'napari_micromanager._gui_objects._mda_widget.MultiDWidget'>
        self._mmc = <CMMCorePlus at 0x7f538092eb10>
    371 if isinstance(wdg, PropertyBrowser):
    372     wdg.setSizePolicy(
    373         QSizePolicy.Policy.Expanding, QSizePolicy.Policy.Expanding
    374     )

File ~/Documents/oss/micro/napari-micromanager/src/napari_micromanager/_gui_objects/_mda_widget.py:23, in MultiDWidget.__init__(self=<napari_micromanager._gui_objects._mda_widget.MultiDWidget(0x55f387b3b110)>, parent=<napari_micromanager.main_window.MainWindow(0x55f386059e80)>, mmcore=<CMMCorePlus>)
     20 def __init__(
     21     self, *, parent: QWidget | None = None, mmcore: CMMCorePlus | None = None
     22 ) -> None:
---> 23     super().__init__(include_run_button=True, parent=parent, mmcore=mmcore)
        parent = <napari_micromanager.main_window.MainWindow(0x55f386059e80) at 0x7f538811a7c0>
[...]

Issue #122, opened 2023-01-25
task: Add test that no widgets modify core state


ianhi:
### Description

Following up from: https://github.com/pymmcore-plus/pymmcore-widgets/pull/121#discussion_r1086611750

Widgets should fundamentally provide a view (and maybe controls) onto the core state. 
Thus we should add a test that:

1. Loads the demo config
2. Records the full state of the core
3. Creates every possible widget
4. Checks that the core state is unchanged

PR #159, opened 2023-07-27
fix: don't calculate time before subclasses have been inited

ianhi:
This fixes the bug that lead to #158. 

this was causing the calc time function to be called before subclasses finished their `init` which could lead to errors. It's also not obvious that we need to call this every time the config is reloaded.

I also removed the setting of the channel group as I'm pretty sure that that does not belong this deep in a widget, and it also seems to violate the spirit of https://github.com/pymmcore-plus/pymmcore-widgets/issues/122. But @fdrgsp I would really appreciate you taking a look to make sure I'm not messing up something else somewhere by removing that.


fdrgsp:
@ianhi looks good to me!

PR #168, opened 2023-07-29
refactor: remove time estimates from MDAWidget

tlambert03:
This PR removes the time estimating functionality from the MDAWidget.  After #157, it became clear that that was a major slowdown.  

https://github.com/pymmcore-plus/useq-schema/pull/117 adds a mechanism to estimate the time of an experiment that is many orders of magnitude faster than the present approach.  Since it will have effectively no overlap with the current code, this PR just deletes that code, and a followup will replace it


tlambert03:
@fdrgsp are you ok with this plan?

fdrgsp (approved):
Good! So in a next PR we will add back `_MDATimeLabel`, all the widgets connection, etc, right?

tlambert03:
> So in a next PR we will add back _MDATimeLabel, all the widgets connection, etc, right?

In some capacity :)

Issue #167, opened 2023-07-29
MDAWidget should be able to load/create MDASequence without core


tlambert03:
I *think* this should be possible, but let me know @fdrgsp if you disagree.

It would be nice if all of the widgets that input/output generic schema objects from useq-schema could do that without having a system configuration loaded in core.  For example, it seems like you should be able to use `MDAWidget.get_state()` to create an MDASequence object even without the core being loaded (imagine a user not on the microscope creating an experiment and sending a json or yaml file to some other computer).  Similarly, it should be able to load the state of a Sequence without having a full core attached, but currently, you get something like this:

```pytb
In [4]: wdg = MDAWidget()

In [5]: wdg.set_state(mda)
/Users/talley/dev/self/pymmcore-widgets/src/pymmcore_widgets/_mda/_mda_widget.py:314: UserWarning: 'DAPI' config or its group doesn't exist in the 'Channel' ChannelGroup!
  self.channel_widget.set_state([c.dict() for c in state.channels])
/Users/talley/dev/self/pymmcore-widgets/src/pymmcore_widgets/_mda/_mda_widget.py:314: UserWarning: 'FITC' config or its group doesn't exist in the 'Channel' ChannelGroup!
  self.channel_widget.set_state([c.dict() for c in state.channels])
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[5], line 1
----> 1 wdg.set_state(mda)

File ~/dev/self/pymmcore-widgets/src/pymmcore_widgets/_mda/_mda_widget.py:335, in MDAWidget.set_state(self, state)
    333 if state.stage_positions:
    334     self.p_cbox.setChecked(True)
--> 335     self.position_widget.set_state(list(state.stage_positions))
    336 else:
    337     self.p_cbox.setChecked(False)

File ~/dev/self/pymmcore-widgets/src/pymmcore_widgets/_mda/_positions_table_widget.py:723, in PositionTable.set_state(self, positions, clear)
    720     raise TypeError("The 'positions' arguments has to be a 'Sequence' type.")
    722 if not self._mmc.getXYStageDevice() and not self._mmc.getFocusDevice():
--> 723     raise ValueError("No XY and Z Stage devices loaded.")
    725 # variables used for autofocus devaice name check
    726 rows = set(range(self._table.rowCount()))

ValueError: No XY and Z Stage devices loaded.
```

fdrgsp:
[...]

PR #169, opened 2023-07-29
fix: remove set_state core link  in PositionTable

fdrgsp:
As discussed in https://github.com/pymmcore-plus/pymmcore-widgets/issues/167, this PR allows to set the state of the MDAWidget even if core stages are not loaded.

example:
```py
from useq import MDASequence
from pymmcore_widgets import MDAWidget
mmc = CMMCorePlus.instance()
mmc.loadSystemConfiguration()
mmc.unloadDevice("XY")
mmc.unloadDevice("Z")
m = MDAWidget()
m.show()
mda = MDASequence(stage_positions=[(1,2,3), (4,5,6)])
m.set_state(mda)
```

closes https://github.com/pymmcore-plus/pymmcore-widgets/issues/167


src/pymmcore_widgets/_mda/_positions_table_widget.py
```diff
@@ -384,7 +384,7 @@ def _add_table_row(
         row: int | None = None,
     ) -> None:
         if not self._mmc.getXYStageDevice() and not self._mmc.getFocusDevice():
-            raise ValueError("No XY and Z Stage devices selected.")
+            return
```
tlambert03, line 387: i find this behavior (do nothing and warn nothing) a little jarring.  Why not just add a position at (0,0,0) rather than aborting altogether?

fdrgsp: That's because since no xy and z stage are loaded, there are no columns visible in the position widget. So even If we add (0,0,0) there won't be anything in the table. If we want to allow this, I think  we need to completely remove the hide/show behavior of the widget. Doe you think it make sense?

tlambert03: And why do we hide stuff?  When is it useful?

fdrgsp: It's never useful I guess:joy: ...I did to that way because I liked to not see a stage column if there is no stage. But if we decide that it is more a problem than a nice feature we can remove it.

tlambert03: Ah, yep remove it. It's too surprising and specific

tlambert03: yeah, looking again now I see this all came along with the autofocus stuff in #147 ... another good example of why large PRs are tough to review.  I would go ahead and basically remove everything from #147 that wasn't strictly necessary to accomplish the collection of autofocus information.  I think that means getting rid of `_on_sys_cfg_loaded`, `_on_property_changed`, and generally all of the "core-following" stuff.

This seems to come up often: the desire to make one master widget that automagically does all the things that (e.g.) napari manager might need it to do when actively connected to a microscope session.  But I think this is probably better accomplished by subclasses for specific scenarios.  Don't do it in this PR, but after we remove this stuff, I'd like to more generally look at how deeply we've automagically connected a live core to our widgets, and consider a global concept of "follow-core".  Certainly, that's the point of many of the widgets, but it probably isn't the point of the all the widgets (like the MDA widgets) and might be better broken up into purpose-driven subclasses

fdrgsp: @tlambert03 I think I removed the column hiding behavior and left only the autofocus stuff the I have added in #147. And I also think that with this changes, the core is not much involved anymore, only in few places where I think it is necessary. But you can have a look and we can make fixes in another PR.

tlambert03:
thanks!  This is great.

In a follow up PR, let's make it so that you can edit the position table (with real numbers) even if there isn't an active stage connected

PR #180, opened 2023-08-06
feat: new useq-schema widgets

tlambert03:
this is a rewrite of the useq schema / MDA widgets.  They contain no core reference, that should come in a follow up.  They have been reconsidered to reduce code duplication, improve readability, clarify the public APIs, and make it possible to work directly with useq-schema objects without needing a core

Screen Shot 2023-08-06 at 3 24 32 PM
Screen Shot 2023-08-06 at 3 24 35 PM
Screen Shot 2023-08-06 at 3 24 37 PM
Screen Shot 2023-08-06 at 3 24 38 PM
Screen Shot 2023-08-06 at 3 24 34 PM


tlambert03:
OK, this is in pretty good shape.  It's not connected to anything yet, but connection to a core object and replacement of the existing mda widget can come in a follow up PR

PR #185, opened 2023-08-29
fix: remove groupbox from config widget

fdrgsp:
This PR removes the `QGroupbox` from the `ConfigurationWidget` so that most basic widget are the same (clean and simple). If a `QGroupbox` is needed, the can be added where the widget is used.

note: napari-micromanager are failing because of this line: https://github.com/pymmcore-plus/napari-micromanager/blob/6abd0f50cf7840b7a08b28bd0ad9a428c484f284/src/napari_micromanager/_gui_objects/_toolbar.py#L143 where we set the `QGroupBox` text. Since there is not a `QGroupBox`, we get the error.

### DIFF
--- a/src/pymmcore_widgets/_load_system_cfg_widget.py
+++ b/src/pymmcore_widgets/_load_system_cfg_widget.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 from pymmcore_plus import CMMCorePlus
 from qtpy.QtWidgets import (
     QFileDialog,
-    QGroupBox,
     QHBoxLayout,
     QLineEdit,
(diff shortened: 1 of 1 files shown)


tlambert03:
are there no places elsewhere in the code where you want to put this into a group box?

tlambert03:
nope, looks like this isn't used internally

tlambert03:
hey @fdrgsp, I didn't notice that this broke the napari-micromanager tests... can you resolve that?
https://github.com/pymmcore-plus/pymmcore-widgets/actions/runs/6014044259/job/16312877421

PR #187, opened 2023-08-31
feat: Stack viewer

wl-stepp:
A viewer for acquisitions run from pymmcore-plus. Let me know your thoughts, happy to adjust so that it fits into the pymmcore-widgets 'philosophy'.


src/pymmcore_widgets/_mda/_stack_viewer.py
```diff
@@ -0,0 +1,284 @@
...
+        time_plan={"interval": 0.5, "loops": 20},
+        axis_order="tpcz",
+    )
+
+    datastore = QLocalDataStore([40, 1, 2, size, size], mmcore=mmcore)
+    w = StackViewer(sequence=sequence, mmcore=mmcore, datastore=datastore)
+    w.show()
+
+    mmcore.run_mda(sequence)
+    qapp.exec_()
```
tlambert03, line 284: can you move this block to a file in `examples`?

src/pymmcore_widgets/_mda/_util/_labeled_slider.py
```diff
@@ -0,0 +1,127 @@
...
+            setattr(self, function, func)
+
+        self.current_value = QtWidgets.QLabel()
+        self.current_value.setText("0")
+        self.current_value.setAlignment(QtCore.Qt.AlignLeft)
+        self.current_value.setSizePolicy(
+            QtWidgets.QSizePolicy.Fixed, QtWidgets.QSizePolicy.Fixed
+        )
+
+        self.play_btn = QtWidgets.QPushButton("▶")
```
tlambert03, line 47: ultimately, i'd like to use a font icon rather than a UTF8 glyph here.  see here for an example:

https://github.com/pymmcore-plus/pymmcore-widgets/blob/54c240c2e41f5fb27dfb0a085e22f2215def6d18/src/pymmcore_widgets/_snap_button_widget.py#L74

`MDI6.play` and `MDI6.stop` would be a good match I suppose

src/pymmcore_widgets/_mda/_util/_labeled_slider.py
```diff
@@ -0,0 +1,127 @@
+from qtpy import QtCore, QtGui, QtWidgets
+
+
+class QLabeledSlider(QtWidgets.QWidget):
```
tlambert03, line 4: as mentioned elsewhere, would be good if we could use the superqt labeled slider and just add play buttons and stuff to that.  I can have a look at that:

https://pyapp-kit.github.io/superqt/widgets/qlabeledslider/

src/pymmcore_widgets/_mda/_stack_viewer.py
```diff
@@ -0,0 +1,284 @@
+import copy
+
+import numpy as np
+from pymmcore_plus import CMMCorePlus
+from qtpy import QtCore, QtGui, QtWidgets
+from useq import MDASequence
+
+from pymmcore_widgets._mda._datastore import QLocalDataStore
+from pymmcore_widgets._mda._util._channel_row import ChannelRow
```
tlambert03, line 9: if vispy wasn't installed, this import of ChannelRow would cause an `ImportError` without the useful message below.  that needs to be resolved

src/pymmcore_widgets/_mda/_datastore.py
```diff
@@ -0,0 +1,89 @@
...
+import numpy.typing as npt
+from pymmcore_plus import CMMCorePlus
+from qtpy import QtCore, QtWidgets
+from useq import MDAEvent
+
+DIMENSIONS = ["t", "c", "z"]
+
+
+class QLocalDataStore(QtCore.QObject):
+    """DataStore that connects directly to the mmcore frameReady event and saves the data for
```
tlambert03, line 13: one-line PEP-257-style docstrings

src/pymmcore_widgets/_mda/_stack_viewer.py
```diff
@@ -0,0 +1,284 @@
...
+
+        if sequence is not None:
+            dims = [x for x in sequence.sizes.keys() if sequence.sizes[x] > 0]
+        else:
+            dims = DIMENSIONS
+        dims.remove("c")
+        self.sliders = []
+        for dim in dims:
+            slider = LabeledVisibilitySlider(dim, orientation=QtCore.Qt.Horizontal)
+            slider.valueChanged[int, str].connect(self.on_display_timer)
```
tlambert03, line 168: avoid using signal overloads ... they have variable support across different qt backends, and are phased out in qt6

tlambert03 (commented):
thanks a lot @wl-stepp! ... as mentioned elsewhere, I definitely want to have a non-napari stack viewer in here.  This is great.

I'll have a number of comments and requests, probably over a few iterations, so please bear with me :)

one thing that we can get started on is the pre-commit failures.  There are plenty of linting and typing errors that we'll need to resolve eventually (we enforce relatively strict typing rules here).  I'm happy to fix those and push to your branch if you want, or you can take a stab at it if you prefer and ask questions if any of the rules are confusing.
(you can pip install `pre-commit` locally then run `pre-commit run -a` to check as you go)

wl-stepp:
Yeah sure, definitely to be seen as work in progress. Will work on it, thanks for the comments.

src/pymmcore_widgets/_mda/_stack_viewer.py
wl-stepp: Added the try/except to the channel row. We could think about not using vispy there at all, but just QColors or something, then translate to vispy colormaps only in the StackViewer.

src/pymmcore_widgets/_mda/_stack_viewer.py
wl-stepp: Interesting, didn't know, sorry. Will handle that when modifying the slider to use the superqt slider.

src/pymmcore_widgets/_mda/_stack_viewer.py
tlambert03: agreed.  could also consider using `cmap` (https://cmap-docs.readthedocs.io/en/latest/) ... which is a minimal-dependency library I created to be able to use colormaps without assuming a specific platform (works for matplotlib, vispy, napari, etc...)

tlambert03:
@wl-stepp ... in case you missed the PR to your branch (sometimes people miss those), I made a version that uses superqt slider here: https://github.com/wl-stepp/pymmcore-widgets/pull/2

wl-stepp:
Not sure why these two tests keep failing.
Windows fatal exception: access violation
@tlambert03 can you help?

tests/test_stack_viewer.py
```diff
@@ -0,0 +1,39 @@
...
+    datastore = QLocalDataStore(shape=(3, 1, 2, 512, 512), mmcore=mmcore)
+    canvas = StackViewer(mmcore=mmcore, datastore=datastore)
+    canvas.show()
+    qtbot.addWidget(canvas)
+    mmcore.run_mda(sequence)
+    if qtbot:
+        with qtbot.waitSignal(datastore.frame_ready, timeout=5000):
+            pass
+        with qtbot.waitSignal(datastore.frame_ready, timeout=5000):
+            pass
```
tlambert03, line 26: i think just waiting for `events.sequenceFinished` would probably fix it here, but just to be safe, let's also use the non-threaded version (`mmcore.mda.run`) for the test:

```suggestion
    with qtbot.waitSignal(mmcore.mda.events.sequenceFinished):
        mmcore.mda.run(sequence)
```

wl-stepp:
Failing on  tests/test_group_preset_widget.py::test_edit_group

tlambert03:
hey @wl-stepp.  Thanks for continuing to work on this.  I've been thinking about it a lot... here are some general thoughts I have:

- I see this as a very high value and useful feature, so definitely want to get it in, but also want to think carefully about the API and generalizability of it.  (really, a lightweight qt nd stack viewer that doesn't pull in all of napari's dependencies and load time is of general usefulness... and I can also imagine something like it being upstreamed to superqt)
- The datastore pattern is very similar to patterns we already have in napari-micromanager (for viewing data as it comes out of pymmcore-plus) as well as in pymmcore-plus itself, where a similar pattern would be needed for a file writer (https://github.com/pymmcore-plus/pymmcore-plus/pull/29).  It's not dissimilar to the "DataHandler" protocol I mentioned in https://github.com/pymmcore-plus/pymmcore-plus/pull/247.  So, while I'm open to just including it as is in an experimental capacity, it's a likely target for refactoring in the interest of unifying all of these very similar downstream applications.
- The store itself uses a greedy numpy array shaped as large as the experiment, which can lead to [memory issues](https://github.com/pymmcore-plus/napari-micromanager/issues/143).  In napari-micromanager, @ianhi  implemented a [zarr-backed datastore](https://github.com/pymmcore-plus/napari-micromanager/pull/145) that works nicely and grows with the experiment.  Not saying that needs to be a part of this here, but it would be nice to think ahead to how this could be more of a swappable array interface, and less tied to an in-memory numpy array.
- Ultimately, I'd love it if the stack viewer didn't need to be so rigid as to what it expects from the experiment in terms of shape. It could grow as it sees new dimensions and/or indices (that's harder with a numpy-backed store, but conceivable either way...).  And since each new frame comes with the `event.index`, it's easy enough to put it into an ever-growing array.
- Note, I was inspired by the color picker and have a PR at superqt implementing a [full colormap picker](https://github.com/pyapp-kit/superqt/pull/195) and [individual color picker](https://github.com/pyapp-kit/superqt/pull/194) ... we could eventually use those here.

These are all "aspirations" and not requirements for this PR.  But I think what I'd like to do here is put this all into an `experimental` namespace.  Something where you can immediately start using it, but where nobody expects the API to be stable and we can iterate on it.  How does that sound to you?

wl-stepp:
Hey @tlambert03,

Yes, sounds all very reasonable to me. An experimental namespace sounds like a good idea. We might have this in use on a microscope reasonably soon, I'm sure there will be a lot of things that come up then.

>     * The datastore pattern is very similar to patterns we already have in napari-micromanager (for viewing data as it comes out of pymmcore-plus) as well as in pymmcore-plus itself, where a similar pattern would be needed for a file writer ([First pass at MDA writers pymmcore-plus#29](https://github.com/pymmcore-plus/pymmcore-plus/pull/29)).  It's not dissimilar to the "DataHandler" protocol I mentioned in [feat: add listeners_connected utility pymmcore-plus#247](https://github.com/pymmcore-plus/pymmcore-plus/pull/247).  So, while I'm open to just including it as is in an experimental capacity, it's a likely target for refactoring in the interest of unifying all of these very similar downstream applications.
>
>     * The store itself uses a greedy numpy array shaped as large as the experiment, which can lead to [memory issues](https://github.com/pymmcore-plus/napari-micromanager/issues/143).  In napari-micromanager, @ianhi  implemented a [zarr-backed datastore](https://github.com/pymmcore-plus/napari-micromanager/pull/145) that works nicely and grows with the experiment.  Not saying that needs to be a part of this here, but it would be nice to think ahead to how this could be more of a swappable array interface, and less tied to an in-memory numpy array.


Yes, this seemed like the lowest effort quick solution to this. I will have a look at the others and we can try to come up with something more long-term.

>     * Ultimately, I'd love it if the stack viewer didn't need to be so rigid as to what it expects from the experiment in terms of shape. It could grow as it sees new dimensions and/or indices (that's harder with a numpy-backed store, but conceivable either way...).  And since each new frame comes with the `event.index`, it's easy enough to put it into an ever-growing array.

The numpy array grows if an index exceeds the available space (https://github.com/wl-stepp/pymmcore-widgets/blob/a41dfde19673769a89f0780fee2ddf7f001cbb99/src/pymmcore_widgets/_mda/_datastore.py#L79)
But that's very slow for this implementation as a numpy array, something zarr based should be much faster there I expect.

Looking forward to keep working on this together!

tlambert03:
Sounds good!

tlambert03:
hey @wl-stepp, as soon as you want to have this in main under the `experimental` namespace, let me know.  or if there are bits you want to discuss...   
As mentioned, I see this as a feature that we'll keep iterating on, but no reason not to get this in if it's something you are using.

wl-stepp:
Ok, this version runs decent on our scope for now. I think we could put it into an experimental namespace if you want to. I would do another round of edits trying to be less opinionated about the sequence of events coming in towards the end of February. I think that should be possible and worth it now with the zarr storage in the background.
I'm not in the office at the moment, so response times might be a little longer.

tlambert03:
I just pulled it and played with it a bit.  this is an awesome start!

I'm inclined to just merge this as is, and put it in `pymmcore_widgets.experimental`.  I'd like to play around with it a bit, nothing major, perhaps some GUI tweaks and stuff.  How would you like to go about collaborating on this, without messing up your usage of it?  If we merge into main/experimental, is it ok to keep making PRs and tweaking it and having you review?  will you try to keep using the main version or would that risk you ultimately using a fork version that you know hasn't changed?

tlambert03:
one first thought is that I'd like to see how it feels if the channel controls were entirely in a single line, rather than blocks as they currently are:

image

something more like this:

image

I feel like that might extend to more than 3 channels better (it already is a bit pressed for space at 3 channels)

tlambert03:
to be clear, I'm not suggesting that change needs to happen in this PR, just an example of things we could play with.  Let me know if/when you're good to merge and iterate in this repo.

wl-stepp:
> one first thought is that I'd like to see how it feels if the channel controls were entirely in a single line, rather than blocks as they currently are:
> image
> 
> something more like this:
> image
> 
> I feel like that might extend to more than 3 channels better (it already is a bit pressed for space at 3 channels)

Yes, I like the boxes to see which channel is active for the value info text. But in this setup some other indicator would work better I guess.

wl-stepp:
Should be no problem. I'm really only relying on the events for interaction in the software for the scope. So should be pretty robust against changes like this.

tlambert03:
thanks again!  I moved some stuff around, created a new `_stack_viewer` folder (the `_mda` folder is probably going away eventually, it mostly has the old MDA widgets), created an `experimental` module with the `StackViewer`, and did some very minor other things like removed the private import from `superqt.cmap`.  Will merge after tests pass!  thanks so much for getting this going @wl-stepp!

tlambert03:
the sporadic test fails on py3.10 pyside6 are legitimate, I'm able to reproduce them (occasionally) locally in a python 3.10 environment.  Usually, it happens only at the end, after all the tests pass successfully, with errors like:

```
zsh: segmentation fault  pytest tests/test_stack_viewer.py -v -x
zsh: bus error  pytest tests/test_stack_viewer.py
```

but occasionally you'll also get them like this:

```
tests/test_stack_viewer.py .....Fatal Python error: Bus error

Current thread 0x00000001de4d9000 (most recent call first):
  Garbage-collecting
```

wl-stepp:
I was not able to pin it down to one of the tests. If I run them individually I can't reproduce. I do get a garbage collection warning for test_connection_warning running individually.

```
gc:0: ResourceWarning: gc: 1 uncollectable objects at shutdown; use gc.set_debug(gc.DEBUG_UNCOLLECTABLE) to list them
```
Not sure why, but might be connected to signalling here. But it's also not the direct cause for the errors above, as those also occur if I comment out the connection_warning test.

tlambert03:
Yeah I suspect it's an aberrant connection or failed cleanup somewhere.  I'm not terribly worried about it in the short term, but I don't want tests to be failing on main.  So, we could skip the tests for now on the failing platforms on CI, but it would need to be fixed before coming out of experimental

wl-stepp:
Ok, I'll try to put some time aside in the next weeks for adjustments to be more agnostic to the sequence and we'll see if that has an impact already.

wl-stepp:
I get the access violation error when the log file is blocked by another process. Maybe related...

wl-stepp:
Ok, interaction test seems to work with the new strategy to add images only when needed. Now we're back to the segmentation and access violation fails.

wl-stepp:
Ok, not sure where else to look.
With Pyside6 and python 3.10 I get a couple of left open widgets when running locally for different tests:
```
ERROR tests/test_core_state.py::test_core_state_unchanged[MDASequenceWidget] - AssertionError: topLevelWidgets remaining after 'test_core_state.py::test_core_state_unchanged': [<PySide6.QtWidgets.QWidget(0x21e88d27a10) at 0x0000021E87CB9940>]
ERROR tests/test_mda_widget.py::test_mda_widget_load_state - AssertionError: topLevelWidgets remaining after 'test_mda_widget.py::test_mda_widget_load_state': [<PySide6.QtWidgets.QLabel(0x21e9223d500) at 0x0000021E8AF29480>]
ERROR tests/test_mda_widget.py::test_mda_buttons - AssertionError: topLevelWidgets remaining after 'test_mda_widget.py::test_mda_buttons': [<PySide6.QtWidgets.QLabel(0x21e9223d500) at 0x0000021E8AF29480>, <PySide6.QtWidgets.QLabel(0x21e92235ca0) at 0x0000021E8AFB5F80>]
ERROR tests/test_mda_widget.py::test_mda_methods - AssertionError: topLevelWidgets remaining after 'test_mda_widget.py::test_mda_methods': [<PySide6.QtWidgets.QLabel(0x21e91ea0740) at 0x0000021E932860C0>, <PySide6.QtWidgets.QLabel(0x21e9223d500) at 0x0000021E8AF29480>, <PySide6.QtWidgets.QLabel(0...
ERROR tests/test_useq_widgets.py::test_mda_wdg - AssertionError: topLevelWidgets remaining after 'test_useq_widgets.py::test_mda_wdg': [<pymmcore_widgets._property_widget.RangedFloatWidget(0x21ec1353ff0) at 0x0000021E93345700>, <pymmcore_widgets._property_widget.RangedFloatWidget(0x21ec1354650)...
ERROR tests/test_useq_widgets.py::test_mda_wdg_load_save[json] - AssertionError: topLevelWidgets remaining after 'test_useq_widgets.py::test_mda_wdg_load_save': [<pymmcore_widgets._property_widget.RangedFloatWidget(0x21ec1353ff0) at 0x0000021E93345700>, <pymmcore_widgets._property_widget.RangedFloatWidget(0x21...
ERROR tests/test_useq_widgets.py::test_mda_wdg_load_save[foo] - AssertionError: topLevelWidgets remaining after 'test_useq_widgets.py::test_mda_wdg_load_save': [<pymmcore_widgets._property_widget.RangedFloatWidget(0x21ec1353ff0) at 0x0000021E93345700>, <PySide6.QtWidgets.QFrame(0x21f1443e880) at 0x0000021E933...
ERROR tests/test_useq_widgets.py::test_proper_checked_index - AssertionError: topLevelWidgets remaining after 'test_useq_widgets.py::test_proper_checked_index': [<pymmcore_widgets._property_widget.RangedFloatWidget(0x21ec1353ff0) at 0x0000021E93345700>, <PySide6.QtWidgets.QWidget(0x21ed771c010) at 0x0000021...
ERROR tests/test_useq_widgets.py::test_mda_wdg_with_autofocus - AssertionError: topLevelWidgets remaining after 'test_useq_widgets.py::test_mda_wdg_with_autofocus': [<pymmcore_widgets._property_widget.RangedFloatWidget(0x21ec1353ff0) at 0x0000021E93345700>, <PySide6.QtWidgets.QWidget(0x21ebf6e62e0) at 0x00000...
```

In that same environment `pytest -k prop_widget` sometimes crashes while 'pytest -k stack' doesn't seem to...

related: https://blog.xmatthias.com/post/pytest-debug-segfault/

tlambert03:
ok, tests are passing now.  I think most of the problem is coming from the internal connections in StackViewer in the frameReady method (i.e. frameReady -> _redisplay -> _reemit -> frameReady).  That loop seems pretty error prone and likely to work under some conditions but not all, and it looks like pyside6 might be particularly sensitive to it.  Also, testing mouse interactions is always hard, so I removed the `qtbot.mouseMove` calls (which often do fail in a platform specific way) for a more direct approach.  Lastly, I removed as many calls to `qtbot.wait()` as possible.  It's always better to do something a bit more explicit, like wait for a specific signal rather than a time interval.  Here I think the waiting is just providing that frameReady loop enough time to resolve itself. 

I do think we should try to improve what happens on frameReady in subsequent PRs, but since it's in experimental, and since it is useful in its current state, I'm going to just merge as is.  I'd say try not to use the StackViewer programmatically too much for now (just use it via the gui) so as to avoid breaking changes as we refactor it.  

thanks again for the contribution!! super important and super useful.

PR #192, opened 2023-09-04
feat: runnable, core-connecte mda widget, based on new useq widgets

tlambert03:


tlambert03:
@fdrgsp ... this is pretty close as well.  Could you have a look at this compared to the current MDA widget, and bullet point any remaining things you think should be included/fixed/restyled before merge?

src/pymmcore_widgets/mda/_core_mda.py
```diff
@@ -0,0 +1,251 @@
...
+    def setValue(self, value: MDASequence) -> None:
+        """Get the current state of the widget."""
+        super().setValue(value)
+        self.save_info.setValue(value.metadata.get("pymmcore_widgets", {}))
+
+    # ------------------- private API ----------------------
+
+    def _update_channel_groups(self) -> None:
+        ch = self._mmc.getChannelGroup()
+        self.channels.setChannelGroups({ch: self._mmc.getAvailableConfigs(ch)})
```
fdrgsp, line 95: I think it is safer if we do something like this here so that we see the combobox all the time. What do you think?

```suggestion
        self.channels.setChannelGroups(
            {
                group_name: self._mmc.getAvailableConfigs(group_name)
                for group_name in self._mmc.getAvailableConfigGroups()
            }
        )
        if ch := self._mmc.getChannelGroup():
            self.channels._group_combo.setCurrentText(ch)
```

src/pymmcore_widgets/mda/_core_mda.py
```diff
@@ -0,0 +1,251 @@
...
+        layout.addWidget(self.control_btns)
+
+        # ------------ connect signals ------------
+
+        self.control_btns.run_btn.clicked.connect(self._on_run_clicked)
+        self.control_btns.pause_btn.released.connect(self._mmc.mda.toggle_pause)
+        self.control_btns.cancel_btn.released.connect(self._mmc.mda.cancel)
+        self._mmc.mda.events.sequenceStarted.connect(self._on_mda_started)
+        self._mmc.mda.events.sequenceFinished.connect(self._on_mda_finished)
+        self._mmc.events.channelGroupChanged.connect(self._update_channel_groups)
```
fdrgsp, line 75: maybe add `systemConfigurationLoaded` event?
```suggestion
        self._mmc.events.channelGroupChanged.connect(self._update_channel_groups)
        self._mmc.events.systemConfigurationLoaded.connect(self._update_channel_groups)
```

src/pymmcore_widgets/mda/_core_mda.py
tlambert03: i was kinda hoping to avoid the added complexity.  but if `getChannelGroup()` returns nothing, then I agree it would be good to show everything.

tlambert03:
> Failing after 1s — 90.99% of diff hit (target 91.03%)

i'm ok with a miss of 0.04 % :joy:

PR #200, opened 2023-09-22
fix: remove grid paint + fix grid units for GridWidthHeight

fdrgsp:
- removing `_GridRendering` 

- in `useq.GridWidthHeight` the units for `width` and `height` are expressed in `µm` but it can be easier for a user to input them in `mm` when using the GUI. This minor modification converts the units in `µm` or `mm` when needed (a valid alternative is to use µm directly in the GUI).


src/pymmcore_widgets/useq_widgets/_grid.py
```diff
@@ -190,14 +191,7 @@ def __init__(self, parent: QWidget | None = None):
...
         if (val := self.value()) is None:
-            return  # pragma: no cover
-        # make a tmp grid object to pass to the _GridRendering widget
-        if isinstance(val, useq.GridRowsColumns):
-            draw_grid = val.replace(relative_to="top_left")
-            draw_grid.fov_height = 1 / ((draw_grid.rows - 1) or 1)
-            draw_grid.fov_width = 1 / ((draw_grid.columns - 1) or 1)
-            self._grid_img.setGrid(draw_grid)
-
+            return
```
tlambert03, line 194:
```suggestion
            return  # pragma: no cover
```

tlambert03: (or just put the emit inside `is not None`

PR #201, opened 2023-09-22
feat: add back the autofocus logic to the PositionTable

fdrgsp:
add back the `autofocus` logic to the `PositionTable` widget.

MERGE after #211 and #213 

Screenshot 2023-09-27 at 11 42 50 AM


tlambert03:
@fdrgsp, is there any reason we can't use `getAutoFocusDevice()` to determine whether or not to present the option to autofocus?

fdrgsp:
@tlambert03 I think the logic is back to what we were doing in the other widget version. 

The main thing to check are the `value()` and `setValue()` methods of the `PositionTable`. `setValue()` can definitely be spit in smaller methods but can you check if you agree with the logic?

tlambert03:
let me know when it's time to review this again

src/pymmcore_widgets/useq_widgets/_positions.py
```diff
@@ -185,16 +198,63 @@ def value(
         ):
             if not r.get(self.NAME.key, True):
                 r.pop(self.NAME.key, None)
-            out.append(useq.Position(**r))
+
+            if not self.use_af.isChecked():
+                pos = useq.Position(**r)
+            else:
+                pos = useq.Position(**r)
```
tlambert03, line 205: r is a dict of kwargs used to create the position.  rather than creating the position with **r and then fixing it, can you just update the kwargs before creation?

src/pymmcore_widgets/useq_widgets/_positions.py
```diff
@@ -185,16 +198,63 @@ def value(
...
+                    )
+
+            out.append(pos)
         return tuple(out)
 
     def setValue(self, value: Sequence[useq.Position]) -> None:  # type: ignore
         """Set the current value of the table."""
         _values = []
+        _use_af = False
+        # _af_z_devices: set[str] = set()
```
tlambert03, line 229: remove commented out code if unused

src/pymmcore_widgets/useq_widgets/_positions.py
```diff
@@ -185,16 +198,63 @@ def value(
...
+                # get autofocus plan device name and offset
+                _af_offset = v.sequence.autofocus_plan.autofocus_motor_offset
+
+                # set the autofocus offset that will be added to the table
+                _af = {self.AF.key: _af_offset}
+
+                # remopve autofocus plan from sub-sequence
+                v = v.replace(sequence=sub_seq)
+
+                _use_af = True
```
tlambert03, line 252: this logic feels a bit long/convoluted. need to look later when I'm able to run it locally

tlambert03:
couple quick thoughts when running the example:

1. the end result of adding a few positions with the same AF offset is pretty "ouch":

Screen Shot 2023-10-09 at 7 52 20 AM

it would be much nicer if that could end up as:

```yaml
autofocus_plan:
  autofocus_motor_offset: 0.0
  axes:
  - p
stage_positions:
- x: 0.0
  y: 0.0
  z: 0.0
- x: 0.0
  y: 0.0
  z: 0.0
- x: 0.0
  y: 0.0
  z: 0.0
```

I recall having a conversation about that, but can't remember what the conclusion was.

2. `_get_autofocus_axes` feels like a return to magical behavior. (i.e. you're deciding for the user what axes to include rather than letting them explicitly tell you).   This is the same as the comment above, which is still unresolved: https://github.com/pymmcore-plus/pymmcore-widgets/pull/201/files/35183d6391e6284fb436aa6b0ac22757d0ea04d6#diff-99e344cf77fae0afa94cc80e00152db69fc033af876e3bfca224f9e6dc230e17

fdrgsp:
> it would be much nicer if that could end up as:
> 
> ```yaml
> autofocus_plan:
>   autofocus_motor_offset: 0.0
>   axes:
>   - p
> stage_positions:
> - x: 0.0
>   y: 0.0
>   z: 0.0
> - x: 0.0
>   y: 0.0
>   z: 0.0
> - x: 0.0
>   y: 0.0
>   z: 0.0
> ```

Are you suggesting that it might be better to have two different behaviors? One if all the positions have the same af offset and one if at least one position has a different one?



> _get_autofocus_axes feels like a return to magical behavior. (i.e. you're deciding for the user what axes to include rather than letting them explicitly tell you).

👍🏻 I'll try to add a way for the user to select the axis, probably checkboxes can be a good option. Do we agree that  `z_plan` and `channels` should not be an option for the af axis?

tlambert03:
> Are you suggesting that it might be better to have two different behaviors? One if all the positions have the same af offset and one if at least one position has a different one?

yep. check if all the offsets are the same and if so, apply it to the global sequence, otherwise use individual sequences.

tlambert03:
> 👍🏻 I'll try to add a way for the user to select the axis, probably checkboxes can be a good option. Do we agree that  z_plan and channels should not be an option for the af axis?

it would be fine for this PR to just use 'p' as well and add the checkboxes to others in another PR.  and leave that decision for later

tlambert03 (approved):
lovely!

PR #203, opened 2023-09-23
fix: don't set Null MDASequence values in position table, cast dicts to MDASequence in MDAButton.setValue

fdrgsp:
@tlambert03  This PR fixes 2 issues that I found

## fix 1
if we pass a `useq.Position` that has a `sub-sequence` that is not `None` but empty (e.g. `useq.Position(x=1, y=2, z=3, sequence=useq.MDASequence())`) to the `setValue()` method of the `PositionTable`,  we should **_not_** add this as a `sub-sequence` in the widget (and so the sub-sequence cancel button should not be visible). 

Calling the `value() `method after that should give us that the `useq.Position` sequence is `None` (e.g. `Position(x=1.0, y=2.0, z=3.0, name=None, sequence=None)`).

This PR fixes this by checking if the `useq.Position` sequence is empty before appending it in the `setValue() `method.

to reproduce the behavior:
```py
from pymmcore_widgets.useq_widgets import PositionTable
import useq
p = PositionTable()
p.show()
pos = useq.Position(x=1, y=2, z=3, sequence=useq.MDASequence())
p.setValue([pos])
```
After running the code above, without this fix, it will look like the position has a sub-sequence (cancel button is visible in the sub-sequence column) even if the  sub-sequence is empty.

## fix 2
after using the `setValue()` method of the `PositionTable` we assign the `_value` attribute to the `sub-sequence` button as a `dict`. For example: 

```py
pt = PositionTable()
pos = useq.Position(x=1, y=2, z=3, sequence=useq.MDASequence(grid_plan=useq.GridRowsColumns(rows=2, columns=3)))
pt.setValue([pos])
```

If then we click on the `sub-sequence` button, we get a `TypeError` because at the moment `_value` can only be a `useq.MDASequence`.

This PR fixes this in the `_MDAPopup` widget by converting the `_value` into a `useq.MDASequence` if it is a `dict`.

to reproduce the behavior:
```py
from pymmcore_widgets.useq_widgets import PositionTable
import useq
p = PositionTable()
p.show()
pos = useq.Position(x=1, y=2, z=3, sequence=useq.MDASequence(grid_plan=useq.GridRowsColumns(rows=2, columns=3)))
p.setValue([pos])
```
After running the code above, without this second fix, if clicking on the `sub-sequence` button will give a `TypeError`.

I've added two tests for this two fixes.


tlambert03:
good catches!  and thanks for the test.
I made a slight change, putting all of the logic on `MDAButton.setValue`, where it really matters.  
- fix 1: rather than checking in `PositionTable.setValue` let's check a direct comparison against `useq.MDASequence()` in `MDAButton.setValue`
- fix 2: rather than allowing `_MDAPopup` to deal with bad input, let's directly fix the faulty value of `MDAButton` by casting dicts as necessary

PR #208, opened 2023-09-26
fix: update CoreConnectedPositionTable + test

fdrgsp:
This PR update the functionality of the `CoreConnectedPositionTable`:

- when clicking on the add button, we add the current XY or Z stage coordinates.
- handles the cases where the XY or Z stage are not selected/loaded. If not loaded, the respective table column will be hidden and the `value()` method will output `None` for the x, y, and/or z positions coordinates.

to see this behavior:

```py
from pymmcore_widgets.mda._core_positions import CoreConnectedPositionTable
import useq
mmc = CMMCorePlus.instance()
mmc.loadSystemConfiguration()
mmc.unloadDevice("Z")
p = CoreConnectedPositionTable()
p.show()

pos = [useq.Position(x=1, y=2, z=3)]
p.setValue(pos)

print(p.value())
>>> (Position(x=1.0, y=2.0, z=None, name=None, sequence=None),)
```

### DIFF
--- a/src/pymmcore_widgets/mda/_core_positions.py
+++ b/src/pymmcore_widgets/mda/_core_positions.py
@@ -1,14 +1,11 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
 from fonticon_mdi6 import MDI6
 from pymmcore_plus import CMMCorePlus
-from qtpy.QtWidgets import (
-    QCheckBox,
-    QWidget,
-    QWidgetAction,
-)
+from qtpy.QtWidgets import QCheckBox, QWidget, QWidgetAction
+from superqt.utils import signals_blocked
 
 from pymmcore_widgets.useq_widgets import PositionTable
 from pymmcore_widgets.useq_widgets._column_info import ButtonColumn
@@ -46,6 +43,10 @@ class CoreConnectedPositionTable(PositionTable):
         self.table().addColumn(self._xy_btn_col, self.table().indexOf(self.X))
         self.table().addColumn(self._z_btn_col, self.table().indexOf(self.Z) + 1)
 
+        # when a new row is inserted, call _on_rows_inserted
+        # to update the new values from the core position
+        self.table().model().rowsInserted.connect(self._on_rows_inserted)
+
         # add move_to_selection to toolbar and link up callback
         toolbar = self.toolBar()
         action0 = next(x for x in toolbar.children() if isinstance(x, QWidgetAction))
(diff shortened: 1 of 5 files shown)


src/pymmcore_widgets/mda/_core_positions.py
```diff
@@ -51,6 +52,54 @@ def __init__(
...
+    def _on_property_changed(self, device: str, prop: str, value: str) -> None:
+        """Update the autofocus device combo box when the autofocus device changes."""
+        if device != "Core" or prop not in {"XYStage", "Focus"}:
+            return
+        if prop == "XYStage":
+            self._enable_xy()
+        elif prop == "Focus":
+            self._enable_z()
+
+    def _add_row(self) -> None:
```
tlambert03, line 94: this is definitely super important, but it's tricky to implement.

As is currently, the valueChanged signal will be emitted before this update occurs, so anyone listening to valueChanged will get the wrong value for the positions.

Since the default/new values for a new row are provided by the ColumnInfo instance, I think it would be nice to have a `default_factory` option on ColumnInfo that can retrieve a default value as a function.  Then the values would be there *as* they get added to the table, and regardless of how the adding happened (button click or not).  Unfortunately, it would then be hard to pass the core instance to the default factory function.  So it might not be possible to use anything but the global core instance to populate the default values...

need to ponder this

tlambert03: I just pushed some updates to this that should make sure this a) happens regardless of how the new row was added and b) only emits one signal per change

tests/test_useq_core_widgets.py
```diff
@@ -62,14 +67,136 @@ def test_core_connected_position_wdg(qtbot: QtBot, qapp) -> None:
...
     assert round(p0.z) == 33
 
+    wdg._mmc.waitForSystem()
     pos_table.move_to_selection.setChecked(True)
     pos_table.table().selectRow(0)
     pos_table._on_selection_change()
+
+
+@pytest.mark.parametrize("stage", ["XY", "Z"])
+def test_core_connected_position_wdg_cfg_loaded(
```
tlambert03, line 85: these two tests are extremely similar... can you find a way to reduce a bit of the code duplication?

fdrgsp: yep :) I'm working on it

tlambert03:
@fdrgsp ... just pushed an update that removes the value() method override (that removed hidden columns) in favor of an `exclude_hidden_cols` attribute to the `DataTable.iterRecords` method itself.  If True, it omits hidden columns from the values dict

https://github.com/pymmcore-plus/pymmcore-widgets/pull/208/commits/59d1947722d5cc8e32810e7f4471fec221f2a91d

PR #209, opened 2023-09-27
fix: increase FloatColumn min/max in Position table

fdrgsp:
add `minimum` and `maximum` parameters to the `FloatColumn`.


src/pymmcore_widgets/useq_widgets/_positions.py
```diff
@@ -134,10 +134,11 @@ class SubSeqColumn(WidgetColumn):
 class PositionTable(DataTableWidget):
     """Table for editing a list of `useq.Positions`."""
 
+    v_max, v_min = (9999999, -9999999)
```
tlambert03, line 137: don't add these as public class attributes, it might look like the table itself has v_max and v_min attributes that seemingly "do" something if you set them.  Put it outside the class.  And, since it's just a negative one, you can use a single `MAX = 999999` value

PR #211, opened 2023-09-28
fix: use core connected widgets for core connected  sub-sequence _MDAPopup

fdrgsp:
use the `CoreConnectedGridPlanWidget()` and `CoreConnectedZPlanWidget()` when opening sub-sequence `_MDAPopup` widget.

Screenshot 2023-09-28 at 8 31 31 AM

### DIFF
--- a/examples/mda_useq.py
+++ b/examples/mda_useq.py
@@ -1,7 +1,7 @@
 import useq
 from qtpy.QtWidgets import QApplication
 
-from pymmcore_widgets.mda._core_mda import MDASequenceWidget
+from pymmcore_widgets.useq_widgets import MDASequenceWidget
 
 app = QApplication([])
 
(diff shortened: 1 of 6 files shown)


examples/mda_useq.py
```diff
@@ -1,7 +1,7 @@
 import useq
 from qtpy.QtWidgets import QApplication
 
-from pymmcore_widgets.useq_widgets import MDASequenceWidget
```
fdrgsp, line 4: Had to do this for circular import but I don't think is a good idea. Need to fix this

tlambert03:
thanks! I agree this is an important thing to have.  I want to play a bit with different patterns

src/pymmcore_widgets/mda/_core_positions.py
```diff
@@ -20,7 +25,36 @@ class SaveInfo(TypedDict):
         should_save: bool
 
 
+class CoreConnectedMDAButton(MDAButton):
```
tlambert03, line 28: Are you going to remove this stuff ?

tlambert03 (approved):

PR #213, opened 2023-09-28
fix: add position if relative z_plan

fdrgsp:
If building an `MDASequence` with a `relative` `z_plan` without a specified `Position`, we need to specify the relative starting position.

### DIFF
--- a/src/pymmcore_widgets/mda/_core_mda.py
+++ b/src/pymmcore_widgets/mda/_core_mda.py
@@ -18,7 +18,7 @@ from qtpy.QtWidgets import (
     QWidget,
 )
 from superqt.fonticon import icon
-from useq import AxesBasedAF, MDASequence, Position
+from useq import MDASequence, Position
 
 from pymmcore_widgets.useq_widgets import MDASequenceWidget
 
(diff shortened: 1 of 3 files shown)


src/pymmcore_widgets/mda/_core_mda.py
```diff
@@ -88,11 +87,24 @@ def _on_sys_config_loaded(self) -> None:
     def value(self) -> MDASequence:
         """Set the current state of the widget."""
         val = super().value()
+
+        # if the z plan is relative, and there are no stage positions, add the current
+        # stage position as the relative starting one
+        if val.z_plan and val.z_plan.is_relative and not val.stage_positions:
+            val = val.replace(stage_positions=[self._get_current_stage_position()])
```
tlambert03, line 94: just a thought... might this not be better dealt with on the consumer side?  For example, if the mda engine receives an MDAsequence that wants a relative Z plan but provides no stage positions, then *it* would figure out what the current position is.

fdrgsp:
closing. adding to pymmcore-plus

fdrgsp:
Reopening since for now it's best to check this here in the `MDAWidget` but when we will introduce the concept of `"relative"` in `useq-schema` we should move it somewhere else.

src/pymmcore_widgets/mda/_core_mda.py
```diff
@@ -88,11 +87,24 @@ def _on_sys_config_loaded(self) -> None:
     def value(self) -> MDASequence:
         """Set the current state of the widget."""
         val = super().value()
+
+        # if the z plan is relative, and there are no stage positions, add the current
```
tlambert03, line 91: Can you add a comment here that this would ideally not be the final solution here.  Explaining that it would be better to put it on the runner somehow

PR #216, opened 2023-10-02
fix: disable x, y, z set buttons when 'move to' is checked

fdrgsp:
@tlambert03 not sure what you think about it but I wanted to make more evident the the state of the `PositionTable` when the `move_to_selection` `QCheckBox` is toggled. As you can see in the video below, I thought it can be a good idea to hide the "set" and the "sub-sequence" buttons and set the x, y, z `TableDoubleSpinBox` to `readOnly` so that the user does not get confused between moving or setting a position.

https://github.com/pymmcore-plus/pymmcore-widgets/assets/70725613/7bf55506-da65-4e4e-b738-fdc79e348215

### DIFF
--- a/src/pymmcore_widgets/mda/_core_positions.py
+++ b/src/pymmcore_widgets/mda/_core_positions.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, cast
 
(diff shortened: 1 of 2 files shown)


tlambert03:
Eh, I like the ability in elements to be able to click to go to a position, move the stage, and then just add the new position without switching modes.  I think it would be better just to make the buttons not "select" the row as discussed before.  Fixing selection is the solution here in my mind, not working around it

PR #222, opened 2023-10-12
fix: extend exposure widget range

tlambert03:
fixes the issue seen by @raulbola in https://github.com/pymmcore-plus/napari-micromanager/issues/293#issuecomment-1759457236

you should be able to set an exposure below 1ms


raulbola:
Hi @tlambert03 I just realized that the same fix must be made at the MDA, when selecting the channel the minimum exposure is limited to 1ms as well.

tlambert03:
ah yes.  thanks!

raulbola:
Moreover, talking about the MDA, is it possible to run an acquisition without specifing a particular channel. Let's say I just created a .cfg file with a camera and my motorized microscope (I have not defined channels, nothing else), and I want to acquire a simple sequence of 1000 images.

tlambert03:
if it's not currently possible (and i believe it's not), it will be soon.

The MDAWidget has recently undergone a full rewrite (#180 and #192) and work is underway to use the new ones in napari-micromanager (https://github.com/pymmcore-plus/napari-micromanager/pull/292) ... We'll make sure it's possible to run an MDA without any channels in the process.  (will create a new issue).

tlambert03:
> Hi @tlambert03 I just realized that the same fix must be made at the MDA, when selecting the channel the minimum exposure is limited to 1ms as well.

@raulbola ... try again with v0.5.3

raulbola:
> if it's not currently possible (and i believe it's not), it will be soon.
> 
> The MDAWidget has recently undergone a full rewrite (#180 and #192) and work is underway to use the new ones in napari-micromanager ([pymmcore-plus/napari-micromanager#292](https://github.com/pymmcore-plus/napari-micromanager/pull/292)) ... We'll make sure it's possible to run an MDA without any channels in the process. (will create a new issue).

Hi @tlambert03 , I have been testing the MDA and i found some things:
- Actually the MDA can record a time or zstack sequence WITHOUT specifying a particular channel. For example if you load the default MM config file **MMConfig_demo.cfg**, and you open the MDA, the RUN button is clickable and you can indeed run sequences without specifying the channel.
- However, when using the provided **test_config.cfg**, the RUN button becomes not clickable and you have to select a particular channel.
Once you select the channel, and the time interval, the RUN button becomes clickable again.  But here is what seems weird to me... if you now remove the channel and unselect the Channels tab, the RUN button becomes clickable again (even though you have not selected any channel).

Maybe I am missing something, but I though it was worth sharing that with you.

fdrgsp:
@raulbola you are right! This is not an ideal behavior. But as @tlambert03 mentioned above, soon we will update `napari-micromanager` with the new  `MDAWidget` ([pymmcore-plus/napari-micromanager#292](https://github.com/pymmcore-plus/napari-micromanager/pull/292)) and this issue should be fixed.

raulbola:
Hi @fdrgsp , thank you very much for your quick response. 
Sorry if I'm suddenly being very active with the issue, actually I'm very excited about the project.
I was afraid that these problems were due to misuse of napari-micromanager or poor definition of the configuration files.

If there is something I can help with, please let me know.

fdrgsp:
> Sorry if I'm suddenly being very active with the issue, actually I'm very excited about the project.

@raulbola that's awesome you started using it! And feedbacks are always great! So we can fix bugs and improve the code!

PR #238, opened 2023-10-31
docs: update the docs

fdrgsp:


tlambert03:
as discussed... let's pull out the renaming into one PR, then follow up with docs

PR #241, opened 2023-10-31
feat: add step() and setStep() methods to StageWidget

### LINKED ISSUE #240: Pass step size to a StageWidget?
When creating a new StageWidget for a custom GUI, is there a way to pass in a step size to override the default? We have a piezo stage that I would like to set the default to something smaller than 10 (0.1 specifically). I looked at the docs and the function docstring, but didn't see anything immediately.

Or maybe the preferred way to do this to create the Widget then set the step size? If so, can you point me in the right direction to do this when using just a StageWidget?

Thanks!

tlambert03:
closes #240 

@dpshepherd, after this, it would be:

```python
wdg = StageWidget('deviceLabel', step=1)
# or later
wdg.setStep(...)
```

### DIFF
--- a/src/pymmcore_widgets/_stage_widget.py
+++ b/src/pymmcore_widgets/_stage_widget.py
@@ -95,6 +95,7 @@ class StageWidget(QWidget):
         self,
         device: str,
         levels: int | None = 2,
+        step: float | None = 10,
         *,
         parent: QWidget | None = None,
         mmcore: CMMCorePlus | None = None,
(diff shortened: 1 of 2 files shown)


tlambert03:
(the couple test fails are upstream and unrelated to this repo)

fdrgsp:
> (the couple test fails are upstream and unrelated to this repo)

@tlambert03  I discovered what is causing this error and I'm working on it

tlambert03:
thanks!!

PR #244, opened 2023-11-16
feat: PixelConfigurationWidget with similar logic as in micromanager

fdrgsp:
**note**: _'tests/test_pixel_config_widget.py::test_pixel_config_wdg_define_configs'_ test **will fail until** https://github.com/pymmcore-plus/pymmcore-plus/pull/287 **is merged and released**.

---

This PR adds the `PixelConfigurationWidget` to create and edit the pixel configurations.

Differently form the already existing `PixelSizeWidget`, which has been renamed in `ObjectivesPixelConfigurationWidget` in this PR, the `PixelConfigurationWidget` allows a user to store in the pixel configuration any of the loaded device properties (`ObjectivesPixelConfigurationWidget` only stores the device pointing to the objective turret).

Also in this PR:
- test for this widget
- docs

---

Screenshot 2023-11-25 at 10 24 59 AM


tlambert03:
i *think* i would change `value(self) -> list[PixelSizePreset]` and `setValue(self, value: Sequence[PixelSizePreset]) -> None:`

tlambert03 (approved):
small  comment on `value()`, but then LGTM

PR #249, opened 2023-11-29
fix: fix overlap in GridPlanWidget

fdrgsp:
The `GridPlanWidget` overlap spinbox value was not correctly expressed in percentage as in `useq-schema`. This PR fixes it + update the tests.

related to (https://github.com/pymmcore-plus/useq-schema/issues/149#issuecomment-1832557352)

@tlambert03 do you want to switch it form 0-100% to 0-1? Should we change it also in `useq-schema`?


tlambert03:
> @tlambert03 do you want to switch it form 0-100% to 0-1? Should we change it also in useq-schema?

let's start by making it correct here, and then decide separately how it should be.  (they should be the same in either case)

tlambert03 (approved):
🎉

PR #251, opened 2023-12-02
test: fix tests for pymmcore 11.1.1.71.0

tlambert03:
This fixes tests after the release of 11.1.1.71.0, which reverted the strict feature flag added in the previous version (that makes setting pre-init properties after initialization a no-op)


tlambert03:
this fixes the primary issue... there's still something funny going on with pyqt6, but i think it's external to this lib

PR #260, opened 2024-01-10
feat: add the possibility to invert the axis in the stage widget

fdrgsp:
This PR adds two checkboxes to the `StageWidget`, `invert X` and `invert Y`, to invert the direction of the stage movements when the arrow buttons are pressed. 

Screenshot 2024-01-10 at 1 56 19 PM


tlambert03 (approved):
thanks!

PR #262, opened 2024-01-24
feat: update Autofocus related methods, warn when 'Set AF Offset per Position' checked but autofocus not engaged

fdrgsp:
This PR  adds/fixes few things that I found out while using one of  Nikon Ti2-E about the `Autofocus` behavior in the `MDAWidget` and in the `CoreConnectedPositionTable`.


tlambert03:
looks good!

PR #265, opened 2024-01-25
fix: fix bugs in  ChannelTable

fdrgsp:
This PR fix two bugs in the `ChannelTable` and create a `CoreConnectedChannelTable` to handle any changes when using a core.

1) the `_group_combo` is "hidden" if there is only one channel group but if you use the `setChannelGroups` method and you set more than one group, the `_group_combo` does not show up. This PR fixes this behavior.

2) before,  we did not get the channel group **from the `_group_combo`** and when adding a channel using the table button, the channel group was set to the default "Channel" even if the channel group selected in the `_group_combo` had a different name (we did not see that before because in the test config the channel group is named "Channel" as the default). This PR fixes this behavior by adding a `_on_rows_inserted` method connected to the table `model().rowsInserted` signal.

3) add a `CoreConnectedChannelTable` to handle cases when a group is deleted or created or a config is loaded.

4) add tests

### DIFF
--- a/src/pymmcore_widgets/mda/_core_channels.py
+++ b/src/pymmcore_widgets/mda/_core_channels.py
@@ -57,16 +57,6 @@ class CoreConnectedChannelTable(ChannelTable):
         if ch_group and ch_group in self.channelGroups():
             self._group_combo.setCurrentText(ch_group)
 
-    # def _update_channel_groups(self) -> None:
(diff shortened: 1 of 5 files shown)


tlambert03:
permanently adding the group combo box sounds good to me, but let's add a label that says "Group:"

tlambert03 (approved):
lovely

PR #267, opened 2024-01-30
fix: better handle enabling and disabling MDAWidget

fdrgsp:
This PR better handles the enabling and disabling of the `MDAWidget` (and adds a test for the fix).

Currently, while an acquisition is running,  the entire `MDAWidget` is disabled and we cannot move through the mda tabs to check what the settings are. This PR fixes this issue by only disabling the tabs checkboxes and the tabs widgets but not the tabs themself. See videos below.

---
BEFORE

https://github.com/pymmcore-plus/pymmcore-widgets/assets/70725613/cb7518fb-57c5-465c-9298-1aec3b62a5f1

AFTER

https://github.com/pymmcore-plus/pymmcore-widgets/assets/70725613/e5b95b6f-5f62-4a5a-b503-3ab0cc6a267b


src/pymmcore_widgets/mda/_core_mda.py
```diff
@@ -185,6 +200,9 @@ def _confirm_af_intentions(self) -> bool:
     def _enable_widgets(self, enable: bool) -> None:
         for child in self.children():
             if child is not self.control_btns and hasattr(child, "setEnabled"):
+                if isinstance(child, CoreMDATabs):
+                    child._enable_tabs(enable)
+                    continue
```
tlambert03, line 205: rather than putting this `if` clause inside of the outer if clause, and then calling `continue`, just use if/else:

```python
            if isinstance(child, CoreMDATabs):
                child._enable_tabs(enable)
            elif child is not self.control_btns and hasattr(child, "setEnabled"):
                child.setEnabled(enable)
```

PR #273, opened 2024-02-05
fix: choose to update ImagePreview when mda is running

fdrgsp:
Necessary after https://github.com/pymmcore-plus/pymmcore-plus/pull/314

When a MDA is running, we should be able to choose whether the ImagePreview  widget should update or not.

### DIFF
--- a/src/pymmcore_widgets/_image_widget.py
+++ b/src/pymmcore_widgets/_image_widget.py
@@ -18,8 +18,8 @@ _DEFAULT_WAIT = 10
 class ImagePreview(QWidget):
     """A Widget that displays the last image snapped by active core.
 
-    This widget will automatically update when the active core snaps an image
-    or when the active core starts streaming.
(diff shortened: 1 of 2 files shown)


tlambert03:
for what it's worth, i think this whole thing is a bit tricky, and we're wading into "magical" territory.  I *do* think that the correct place for the `imageSnapped` signal is in the `snapImage` method.  But whether a naive `ImagePreview` widget like this should stop updating while MDA is running or not feels to be something that the end user might rather decide.  

I think it makes more sense in napari-micromanager, where we're sort of in control of everything, to make that decision on behalf of the user, and not have *two* updates happening every time an image is snapped in an MDA in the napari window.  But here, where it's unclear whether there is another image-previewing window running, i think it could come as a bit more "surprising" to a user to see that the image suddenly stops updating.  

I think we should consider having a public mechanism to toggle this behavior here

tlambert03 (approved):

PR #274, opened 2024-02-05
fix: update few MDAWidget issues

fdrgsp:
This PR fix few minor issue with the MDAWidget.

# z widget
- change color of `ZPlanWidget` icons that are too light and sometimes not very visible and change scale
- add a spacer line between the `ZPlanWidget` mode buttons and the rest of the widget
# core position table
- "Move to position" behavior: disable autofocus before fullfocus() and re-enable it if it was engaged.
# mda sequence widget
- add "Hardware" to the "Use Autofocus on Axis:" checkbox text ("Use Hardware  Autofocus on Axis:").
- remove focus on save and load buttons. (since you can trigger them by pressing the Return key even if you did not mean to)
# core mda widget
- if no position is provided, use the current one.
- if a user tries to setup a multi-position with relative z stack and unchecks the "Include Z" checkbox, use the current z as the starting point.
- remove focus on run, cancel and pause buttons. (since you can trigger them by pressing the Return key even if you did not mean to)
- remove focus on browse_btn in the saving group
# group preset widget
- remove focus on save and load btns

### DIFF
--- a/src/pymmcore_widgets/mda/_core_positions.py
+++ b/src/pymmcore_widgets/mda/_core_positions.py
@@ -232,19 +232,19 @@ class CoreConnectedPositionTable(PositionTable):
                 )
                 try:
                     self._mmc.enableContinuousFocus(False)
-                    self._perform_autofocus(af_engaged)
+                    self._perform_autofocus()
+                    self._mmc.enableContinuousFocus(af_engaged)
+                    self._mmc.waitForSystem()
                 except RuntimeError as e:
                     logger.warning("Hardware autofocus failed. %s", e)
 
             self._mmc.waitForSystem()
 
-    def _perform_autofocus(self, af_engaged: bool) -> None:
+    def _perform_autofocus(self) -> None:
         # run autofocus (run 3 times in case it fails)
         @retry(exceptions=RuntimeError, tries=3, logger=logger.warning)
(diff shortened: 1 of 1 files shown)


src/pymmcore_widgets/mda/_core_mda.py
```diff
@@ -161,8 +170,28 @@ def _on_run_clicked(self) -> None:
...
+        if (
+            sequence.z_plan
+            and sequence.z_plan.is_relative
+            and len(sequence.stage_positions) > 1
+            and not self.stage_positions.include_z.isChecked()
+        ):
+            QMessageBox.critical(
+                self,
+                "Z Plan Error",
+                "Cannot run a relative Z plan without including a starting Z position. "
```
tlambert03, line 187: generally speaking, in useq-schema, the absence of a value (such as a Z position) is to be interpreted as "use the current value".  Why not use the current z position here?

fdrgsp: True, I've updated the value method with this logic.

tlambert03:
ready for review?

fdrgsp:
> ready for review?

Yes you can have a look 👀 thanks!

tlambert03 (approved):

PR #277, opened 2024-02-06
fix: fix docs not grabbing widget png correctly

fdrgsp:
@tlambert03 this fixes the fact that the png of some widgets in the docs are not correct. We just needed to exclude `QMenu` together with  `QFrame` objects.

### DIFF
--- a/docs/_gen_widget_pages.py
+++ b/docs/_gen_widget_pages.py
@@ -51,12 +51,12 @@ def _example_screenshot(cls_name: str, dest: str) -> None:
 
     app = QApplication.instance() or QApplication([])
     new = [w for w in app.topLevelWidgets() if id(w) not in SEEN]
+    # remove all classes that are Qframe or QMenu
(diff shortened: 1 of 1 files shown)


tlambert03:
thanks!  
btw, I keep forgetting to change these docs fix titles in time.  but in the future, preface changes to documentation with `docs:` ... not `fix:`.  
`fix` implies you fixed a bug in the program...

tlambert03:
not sure if this fixed it afterall?

https://pymmcore-plus.github.io/pymmcore-widgets/widgets/MDAWidget/

fdrgsp:
> not sure if this fixed it afterall?
> 
> https://pymmcore-plus.github.io/pymmcore-widgets/widgets/MDAWidget/

I"m not sure why, locally with pyside6 works all the time 🤷‍♂️...while running `mkdocs serve/build`, if you print the `widget` variable in the `_example_screenshot` method, the widget is always correct...

PR #279, opened 2024-02-16
feat: add pymmcore-plus writer to MDAWidget

fdrgsp:
This PR adds the `pymmcore-plus` writers to the `MDAWidget` by passing a `path` to the `run_mda` method when the `run_button` is clicked.

it includes:
- update the save group so that we can select which writer to use with a combo box
- add the extension ".ome.zarr" or ".ome.tiff" depending on which writer is selected
- construct and update the path before running the sequence so that a unique path is created
- update the filename in the save group box with 3 digits depending on the next available path in the saving folder when the sequence ends
- add tests

note: `napari-micromanager` tests are failing because of the new updates here, in this PR https://github.com/pymmcore-plus/napari-micromanager/pull/315 I removed the old saving logic in favor of this new one.

---

Screenshot 2024-03-02 at 4 02 28 PM


tlambert03:
the filename updating is an important feature, but it's tricky.  If someone says "please save this experiment at this directory with this filename `test.ome.tiff`" ... and then that file doesn't exist (because you've changed it to `test_000.ome.tif`) then that's potentially problematic.  So this needs to be handled with some amount of careful consideration

src/pymmcore_widgets/mda/_core_mda.py
```diff
@@ -197,9 +222,29 @@ def _on_run_clicked(self) -> None:
...
+        metadata = sequence.metadata.get("pymmcore_widgets", None)
+        save_dir = metadata.get("save_dir") if metadata else None
+        save_name = metadata.get("save_name") if metadata else EXP
+        extension = metadata.get("extension") if metadata else None
+
+        # create the writers path and make sure they are unique
+        writer_path = None
+        if save_dir and extension is not None:
+            path = Path(save_dir) / f"{save_name}"
+            writer_path = ensure_unique(path, extension)
```
tlambert03, line 237: because `ensure_unique` is such an opinionated function (i.e. it makes specific decisions that a user may very well want to modify), we need to make it easier to customize it.  Currently the only way they can do that is to either try to monkeypatch a private function (`_util.ensure_unique`), or subclass `MDAWidget` and re-implement a complicated private method `MDAWidget._on_run_clicked`.  So, try to think about providing a nicer way for end users to customize this behavior... even if that end-user is napari-micromanager for example.  Think about exposing very small bits of public functionality that would let someone do this themselves.  Think about what the inputs to that functionality should be and what the outputs should be.

src/pymmcore_widgets/mda/_core_mda.py
```diff
@@ -197,9 +222,29 @@ def _on_run_clicked(self) -> None:
...
+            self._update_save_name_text(writer_path, extension)
+
         # run the MDA experiment asynchronously
-        self._mmc.run_mda(self.value())
-        return
+        self._mmc.run_mda(sequence, output=writer_path or None)  # type: ignore
```
tlambert03, line 241: definitely don't type ignore an unexpected keyword argument!  this is alerting you to a very real problem.  you have to either update the requirement pinning, or check the version and provide a fallback

src/pymmcore_widgets/mda/_core_mda.py
fdrgsp: thanks @tlambert03!

do you mean something like a defined protocol with for example the `ensure_unique` and `get_next_available_path` methods (of course names can be different)

```py
class EnsureUniquePath(Protocol):

    @abstractmethod
    def ensure_unique(self, path: Path | str, extension: str = ".tif", ndigits: int = 3) -> Path:
        ...

    @abstractmethod
    def get_next_available_path(self, path: Path | str, extension: str, ndigits: int = 3) -> Path:
        ...
```

and then in the `MDAWidget` we can have either a method or a property to set the `EnsureUniquePath`

```py
class MDAWidget(MDASequenceWidget):
    def __init__(
        self, *, parent: QWidget | None = None, mmcore: CMMCorePlus | None = None
    ) -> None:
        ...
        self._ensure_unique: EnsureUniquePath = EnsureUniquePath()
        ...

    @property
    def unique_path(self) -> EnsureUniquePath:
        return self._ensure_unique

    @unique_path.setter
    def unique_path(self, value: EnsureUniquePath) -> None:
        self._ensure_unique = value
```

src/pymmcore_widgets/mda/_core_mda.py
fdrgsp: Thanks I meant to do it but I forgot

src/pymmcore_widgets/mda/_core_mda.py
tlambert03: by protocol, I don't necessarily mean it _has_ to be a literally `typing.Protocol`, unless you think the complexity of the problem deserves it for example, do you think you widgets would use *both* `ensure_unique` and `get_next_available_path`?  or is it more likely that `get_next_available_path` is just an implementation detail used by `ensure_unique?`.  If so, then perhaps it's just a single callable function.  Or, it could all be put into a public method that someone could subclass and override.

I guess I'm just saying give it a shot: come up with a proposal to make this customizeable somehow, but in as minimal of a way as possible that befits the complexity of the problem

fdrgsp: @tlambert03 not sure if it's the best way but I created a public the method `get_next_available_paths` that returns the next _**two**_ available path so we can use the first for the `run_mda` and the second to update the `save_name` text in the `MDAWidget`.
A user can override this method and and as long as the return type is a `list` of `Path`, it will work (even if there is only one path in the list. the only thing is that the `save_name` won't be updated ).

fdrgsp: I was thinking that maybe it is even worth making public the `_on_run_clicked` method or `MDASequenceWidget` and `MDAWidget`...

tlambert03:
just adding observations here.  Here's how elements handles this (and I kind of like it).  Rather than radioboxes, they have a combobox to the right of the "Filename" field, that lets you pick between "ND2", "Multipage TIFF", and "Single Tiffs".  When you select "Single Tiffs", the label changes from "Filename:" to "Subfolder" (which I like), and the extension is removed from the name.  Here are some other behaviors:
- if the first file/subfolder is called `img`, then the next are called `img001`, `img002`, and so forth
- if the first file/subfolder is called `img01`, then the next are called `img01001`, `img01002`, and so forth. :joy:
- if you force the single-tiffs subfolder name to have an extension "img.tif", then the first folder will be named `img.tif`, and the next ones are `img001`, `img002`
- if you start with  single-tiffs subfolder named `img.a.b.c`, then the next are `img.a.b001`, `img.a001`, `img001`, :joy:
- when incrementing names, it does go for the next maximum number, skipping any "gaps".  (meaning, if you delete file 002, but leave 003 and 004, then the next one will be 005, not 002)
- if you use the name of an existing file (let's say `img.nd2`), then when you click the run button, it will automatically increment the filename to `img001.nd2` *without* prompting the user
- when the experiment is finished (if it was not aborted) the filename is incremented in the gui at that moment.
- storage formats that use extensions will append the extension to *any* string that doesn't end in the proper extension.  So `img.nd2d` will be saved as `img.nd2d.nd2` ... and the next file will be `img.nd2d001.nd2`
- it's quite strict about extensions: it only recognizes `.tif` in the filename, if you call something `img.tiff`, it will save it as `img.tiff.tif`

fdrgsp:
@tlambert03 when you have time can you let me know what you think about this new logic? You can just run the `mda_widget` example to see if you like it. Thanks!!!

I will work on the tests in the meantime.

tlambert03:
ok, love the behavior here now.  Now would just like to see if we can reduce this to something less than a 480 line PR :)
also, ran into a couple issues when running the example.  Specifically, when setting the dropdown to tiff-sequence, and hitting run again, I got:

```pytb
  File "/Users/talley/dev/self/pymmcore-widgets/src/pymmcore_widgets/mda/_core_mda.py", line 223, in _on_run_clicked
    save_path = self._get_saving_path_from_metadata(sequence)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/talley/dev/self/pymmcore-widgets/src/pymmcore_widgets/mda/_core_mda.py", line 233, in _get_saving_path_from_metadata
    extension = next(ext for ext in EXTENSIONS if ext in save_path.name)
```

src/pymmcore_widgets/mda/_save_widget.py
```diff
@@ -0,0 +1,182 @@
...
+        if (
+            not self.isChecked()
+            or not self.save_dir.text()
+            or not self.save_name.text()
+        ):
+            return None
+
+        # returning a str because the MDASequence yaml does not support Path yet
+        # maybe we can add support for Path in the future
+        return str(Path(self.save_dir.text()) / self.save_name.text())
```
tlambert03, line 167: i'm not sure about this change, reducing the value of the of the widget down to a single string.  The dict before was more explicit, and encoded the state of the widget better.   Can you comment on why you reduced it down to a single string?

as for the conversion from string to yaml.  that should only happen when placing the object into a useq.MDASequence.metadata dict.  In other words, don't do it here in this widget just in case some one elsewhere wants to put the output of `widget.value()` into a useq metadata and then later serialize it.  make sense?  that's thinking to specifically about what one might do with the return value from this method

src/pymmcore_widgets/mda/_save_widget.py
```diff
@@ -0,0 +1,182 @@
...
+
+    def setValue(self, value: Path | str) -> None:
```
tlambert03, line 169: this method should also return to the ability to accept a dict, since it better matches the state of the widget.  It can *also* accept a string or path of you want, but that should just split it into the standard dict

tests/test_useq_core_widgets.py
```diff
@@ -9,10 +10,11 @@
 from qtpy.QtCore import QTimer
 from qtpy.QtWidgets import QMessageBox
 
+from pymmcore_widgets._util import _get_next_available_paths
 from pymmcore_widgets.mda import MDAWidget
 from pymmcore_widgets.mda._core_channels import CoreConnectedChannelTable
 from pymmcore_widgets.mda._core_grid import CoreConnectedGridPlanWidget
-from pymmcore_widgets.mda._core_mda import CoreMDATabs
+from pymmcore_widgets.mda._core_mda import CoreMDATabs, SaveAs
```
tlambert03, line 17: i believe `SaveAs` is gone?

tests/test_useq_core_widgets.py
```diff
@@ -9,10 +10,11 @@
 from qtpy.QtCore import QTimer
 from qtpy.QtWidgets import QMessageBox
 
+from pymmcore_widgets._util import _get_next_available_paths
```
tlambert03, line 13: has been renamed

tests/test_useq_core_widgets.py
fdrgsp: yes it is,  I'm working on updating the test 👍🏻

src/pymmcore_widgets/mda/_core_mda.py
```diff
@@ -186,20 +201,40 @@ def _get_current_stage_position(self) -> Position:
 
     def _on_run_clicked(self) -> None:
         """Run the MDA sequence experiment."""
+        # this is in case the user does not press enter after editing the save name.
+        # it has to be called here or it will not be fast enough to update the value
+        # before the run_mda method is called.
+        self.save_info.save_name.editingFinished.emit()
```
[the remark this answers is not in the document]
fdrgsp: `FocusLineEdit` works well all the time...it only doesn't work if you press the `run button` just after typing something in the save_name line edit without pressing enter, or tab....that's why I added this here in the `_on_run_clicked` method.

src/pymmcore_widgets/mda/_save_widget.py
```diff
@@ -0,0 +1,182 @@
...
+            if name.endswith(ext):
+                return ext
+        return ""
+
+    def _get_writer_from_extension(self, extension: str) -> str:
+        """Get the writer from the extension."""
+        for writer, ext in WRITERS.items():
+            for e in ext:
+                if extension == e:
+                    return writer
```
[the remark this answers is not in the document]
fdrgsp: beautiful :)

src/pymmcore_widgets/mda/_save_widget.py
fdrgsp: I thought it was simpler to just pass a string instead of a dict with "save_dir" and  "save_name". Do you think it's better to use the dict? what if we use the `tiff-sequence`, should we change /add a key "subfolder"?

tlambert03: if we use a dict, there's no question about what was in the top field, and what was in the bottom (path) field.  If we use a string, it's up to the receiver to split them.  And in most cases, they'll probably do it correctly, (but still inasmuch as they have to split them anyway, we might as well just give them split).  You could also, for example, include the "requested format" from the combo box in that dict, but if you use a string, the receiver again needs to make a guess based on the extension of that one string.

tlambert03: > what if we use the tiff-sequence, should we change /add a key "subfolder"?

no, don't change the keys in the dict based on their values.  Just keep the keys as whatever they are in the first place (the fact that we change the Label in the GUI is just a convenience for the GUI user, it doesn't need to have any bearing on the fact that this widget is still just a two field widget)

fdrgsp:
@tlambert03 I think you can have another look when you have time.

tlambert03:
thanks!  playing with it now.  Will push some changes shortly, so hang off on updates

tlambert03:
ok @fdrgsp, I'm about done, and you should have another look.  (just need to fix various platform/ci related test failures).
Here are the changes I made:

https://github.com/pymmcore-plus/pymmcore-widgets/pull/279/files/5090738..HEAD

i don't think i changed anything significant.  Reduced some of the tests a bit and refactored other things:
- get_next_available_path should have the same result, perhaps slightly more robust
- reduced the arguments to the public `get_next_available_path` method (the ndigits and extension and stuff is more for the internal method, not something we should expect everyone who wants to override the method to also need)
- made sure that any changes to the save path name are reflected in the widget *while* the file is running (not just afterwards), as well as in the metadata passed downstream to the `run_mda()` method
- allowed save widget `setValue()` to accept a string or Path (but `value()` still returns a dict)
- :eyes: I changed the methods of the save widget a fair amount.  It bothered be a bit how many methods were calling their own `_on_something` methods that are also connected to Qt signals (such as `_on_writer_combo_text_changed` and `_update_combo_text` ... it had a code smell).  So do check whatever it was that caused you to set it up that way and make sure I haven't broken some issue you specifically found.

src/pymmcore_widgets/_util.py
```diff
@@ -114,71 +115,57 @@ def fov_kwargs(core: CMMCorePlus) -> dict:
...
-    stem = path.stem
-
-    # remove digits from the stem if any
-    cur_num = stem.rsplit("_")[-1]
-    if cur_num.isdigit() and len(cur_num) == ndigits:
-        stem = stem[: -ndigits - 1]
-        # if the stem provided has a number, check if path exists and if not, return it
-        new_path = path.parent / f"{stem}_{cur_num}{extension}"
-        if not new_path.exists():
-            return new_path
```
fdrgsp: I think this is the only behavior that is missing but I guess is not important and probably never happens in a normal use.

Basically, if in a `folder` you have a file called `test_001.ome.tif` and the `path` arg in this function is `folder/test_005.ome.tif`, the returned path is `folder/test_005.ome.tif` and not `folder/test_002.ome.tif`.


This is a simple test if you want to see the behavior:

```py
def test_path(tmp_path: Path) -> None:
    (tmp_path / "test_001.ome.tif").touch()
    path = get_next_available_path(tmp_path / "test_005.ome.tif")
    assert path.name == "test_005.ome.tif"
```

fdrgsp (commented):
It looks great!!! Just a small comment but probably not much relevant.

src/pymmcore_widgets/_util.py
tlambert03: ah yep, good catch.  provided that `005` is greater than current_max, i agree.  will fix

tlambert03:
ok! provided tests pass, we'll merge this

tlambert03:
(napari-micromanager fails, at this point, are only because two files will be written)

PR #292, opened 2024-04-24
feat: update the CameraRoiWidget to handle multiple cameras

fdrgsp:
Screenshot 2024-04-24 at 2 40 29 PM


tlambert03:
hey @fdrgsp, is this still something we should get in?  I haven't looked closely yet, but I'm wondering why support for two cameras would require a diff of 700+ lines.  was this harder than it sounds?

fdrgsp:
Hey @tlambert03 , I had to change the logic a bit to support the Multi Camera adapter so that each camera could be controlled individually. I also added a CameraInfo dataclass to store the info of each camera. I did it a while ago so I probably need to dig more into the code to exactly remember why the changes!

PR #295, opened 2024-05-10
fix: update get_next_available_path method

fdrgsp:
if a user is saving a file as `.ome.tiff` from a _multi-position_ acquisition, in the output folder we will have as many file as the number of positions and each file name will have a `_p*` suffix (e.g. `file_p0.ome.tiff`, `file_p1.ome.tiff`, etc.).

At the moment, the `get_next_available_path` method does not consider this `_p*` suffix and if a user doesn't manually change the filename (in the GUI), the new acquisition will overwrite the previous one.

In this PR:
- updated the `get_next_available_path` method with a possible solution to fix the above issue.
- updated the `get_next_available_path` method so that if in the output folder there is already a file with name for example `file_001` and a user sets the save filename for the new acquisition to something not related, for example `test`,  the next available path will not be `test_002` (as it is currently happening) but it will be just `test`.
- added new tests

### DIFF
--- a/src/pymmcore_widgets/_util.py
+++ b/src/pymmcore_widgets/_util.py
@@ -143,17 +143,43 @@ def get_next_available_path(requested_path: Path | str, min_digits: int = 3) -> 
     extension = requested_path.suffix
     # ome files like .ome.tiff or .ome.zarr are special,treated as a single extension
     if (stem := requested_path.stem).endswith(".ome"):
-        extension = ".ome" + extension
+        extension = f".ome{extension}"
         stem = stem[:-4]
+
+    # if ".ome.tif", remove the position index (e.g. _p0, _p1, etc.) from the stem
+    # if any. this can happen when the user is saving a file and ".ome.tif" from a
+    # multi-position acquisition (e.g. "file_001_p0.ome.tif", "file_001_p1.ome.tif",
+    # etc.). we need to remove the _p* to get the pure stem and then (below) check for
+    # existing files to then get the correct counter to add to the stem.
+    if ".ome.tif" in extension:
+        stem = re.sub(r"_p\d+", "", stem)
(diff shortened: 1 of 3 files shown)


src/pymmcore_widgets/_util.py
```diff
@@ -143,17 +143,43 @@ def get_next_available_path(requested_path: Path | str, min_digits: int = 3) ->
     extension = requested_path.suffix
     # ome files like .ome.tiff or .ome.zarr are special,treated as a single extension
     if (stem := requested_path.stem).endswith(".ome"):
-        extension = ".ome" + extension
+        extension = f".ome{extension}"
         stem = stem[:-4]
 
+    # if ".ome.tif", remove the position index (e.g. _p0, _p1, etc.) from the stem
```
tlambert03, line 149: one thing I'm not crazy about with this solution is that we're bleeding an implementation detail of `OMETiffWriter` of `pymmcore_plus.mda.handlers` into this function: specifically, the fact that you may or may not find a `_pN` suffix in the filename is entirely specific to the way that `OMETiffWriter` is (currently) implemented.  That could easily change in the future.  So, we need to determine a better way to encapsulate this information.  It *might* mean that simply using `output=save_path` isn't going to work, but I'm not sure yet.  In any case, we somehow need a more explicit way for the actual writer (which has some control over the filepath) and the widget (which also now has some control over the file path) to communicate about those paths

fdrgsp: That's true...what about if we simply modify the `OMETiffWriter` to create a new folder if the sequence has multiple positions and save each position in this folder? Something similar to this: https://github.com/pymmcore-plus/pymmcore-plus/pull/349.

If we do that, we do not need to update the widget...

PR #305, opened 2024-06-25
[WIP] feat: A vispy-based  stage tracker

fdrgsp:
In this very WIP PR I started thinking about a widget that tracks the stage movements and is able to record the positions (and move to new positions with double-click). The widget is named `StageTracker`. To see an example of usage run the `stage_tracker.py` example.

https://github.com/pymmcore-plus/pymmcore-widgets/assets/70725613/e3cc6ca8-5aa1-4b44-9a30-fe56395d9041

NOTE: there is also a version named `StageRecorder` that is also adding the snapped images in the canvas. This requires more work on how to downsacale the images, etc... To see an example run the `stage_recorder.py` example.


tlambert03:
this is a pretty awesome feature @fdrgsp, and it would be good to get something like this in.  For the sake of review, I think it would be helpful for me to get your thoughts on some high level things, such as what use cases this could serve, and also what the public API of the various objects should be.  I know I can look at the code, but could you also comment here on the main objects and what they do?  Does this build an image map of visited locations (hard to tell from your demo).  Does it work with z positions?  etc... basically, guide us a bit more through what your vision is here, in words

fdrgsp:
closing because of #400

PR #310, opened 2024-07-05
feat: add select all for hub devices

tlambert03:
this PR just adds a Select All checkbox when picking the children of a Hub device that you want to configure

Screenshot 2024-07-05 at 12 15 07 PM


marktsuchida:
Nice, I always want this in MMStudio.

PR #308, opened 2024-07-04
feat: HCS Wizard Widget with useq WellPlatePlan

fdrgsp:
This is a smaller version of #304 that already takes in consideration some of the suggestions and uses the new `useq.WellPlatePlan`.

Here we add the `HCSWizard` but at the moment we cannot update/modify the plate database (the widget is using a default one). In addition, the `HCSWizard` is not yet linked to the `MDAWidget`, it only output a `useq.WellPlatePlan`.

I think it is easier to start from this and add the database and `MDAWidget` link later on.

PR #314, opened 2024-07-07
feat: multi point plan useq widgets

tlambert03:
I'm pulling out the widgets that create multi-points plans from #308, for ease of discussion and testing


tlambert03:
@fdrgsp, you can merge this one after having a look at it, then resolve conflicts with #308.  Note that i made the two widgets regular QWidgets here, so you should insert them into a QGroupBox inside your FOV widget if you want them to have a border

PR #316, opened 2024-07-08
feat: add minimal Points plan view

tlambert03:
borrowed and modified from #308 

hey @fdrgsp, rather than editing each line of the existing FOVSelector, I decided to pull out the bits that I understood and make a bare minimal working widget.  I'm sure that there are some features being missed here, but we can add them back as you point them out (lets be relatively strict about features though).

the main thing I did here was just to allow the GraphicsScene to use real-world coordinates.  the qgraphics framework provides all sorts of conveniences in mapping between coordinate frames of the view and the scene, so to me it seems to make a lot of sense just to place the scene in the "real world" (using microns, just as useq schema positions will use), and then adjust the view as needed.  (pen size is the one annoying thing here, it works as is here, but i think that could also be smarter... need to read more qt docs).  I'll try to add bullet points of things that I know i've removed, and we can add them back, but you can also start a bulleted list as well.

would be great if you could play with this and tell me what you see

Screenshot 2024-07-09 at 4 59 28 PM

-----------

- [x] add back black line path showing the order of acquisition
- [x] discuss the warning that was emitted when the well size didn't match the plate?
- [ ] add back the dotted line that further constrained a random points plan to an area smaller than the well
- [x] there is a minor bug where if you extend the bounds of the scene, you can't "unextend" them for the purpose of autoresizing.  for example, use a grid plan, and add more rows/columns than fit into the coverslip area.  The view will (nicely) resize itself to show the full scene... however, if you then reduce the number of rows/cols, it doesn't "recover" and resize back to the usual size
- [x] deal better with the case of having no `_well_width_um`


fdrgsp:
https://github.com/pymmcore-plus/pymmcore-widgets/blob/4c515835af5ddca78205540040eacd84f13137fb/src/pymmcore_widgets/useq_widgets/points_plans/_points_plan_selector.py#L132-L145

I think at the end here we need to add  `self.valueChanged.emit(self.value())` otherwise, if the radio button is not changing , the `setChecked` method won't trigger the update of the `WellView`...

tlambert03:
added a line 


https://github.com/pymmcore-plus/pymmcore-widgets/assets/1609449/95084203-f427-4ac4-a010-9d64a3ef1352

src/pymmcore_widgets/useq_widgets/points_plans/_random_points_widget.py
```diff
@@ -124,6 +128,7 @@ def value(self) -> RandomPoints:
             order=self.order.currentText(),
             fov_width=fov_x,
             fov_height=fov_y,
+            order=self.order.currentText(),
```
fdrgsp, line 131: is already in line 128
```suggestion
```

src/pymmcore_widgets/hcs/_fov_widget/_well_graphics_view.py
```diff
@@ -0,0 +1,158 @@
...
+
+        # iterate over the plan greedily, catching any warnings
+        # and then alert the model if we didn't get all the points
+        # XXX: I'm not sure about this pattern... feels like the model should be
+        # able to handle this itself, but this is perhaps the only place we actually
+        # iterate over the plan
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore")
+            points = list(plan)
+        if len(points) < plan.num_points:
```
fdrgsp, line 125: this will only work for `useq.RandomPoints` 
```suggestion
        if isinstance(plan, useq.RandomPoints) and len(points) < plan.num_points:
```

tlambert03:
ok, i think this is getting closer.  remaining issues above are

- use random points max_width to further constrain random area, as @fdrgsp had it before
- deal with scene bounds that ratchet up
- deal with (or forbid) `well_width_um = None`
- add tests

i think the first point can come later, the second is also optional.  @fdrgsp, if you want to take a more critical look here at what has been done (knowing we can add more in future PRs) that would be great

src/pymmcore_widgets/hcs/_fov_widget/_well_graphics_view.py
```diff
@@ -0,0 +1,178 @@
...
+        # iterate over the plan greedily, catching any warnings
+        # and then alert the model if we didn't get all the points
+        # XXX: I'm not sure about this pattern... feels like the model should be
+        # able to handle this itself, but this is perhaps the only place we actually
+        # iterate over the plan
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore")
+            points = list(plan)
+        if len(points) < getattr(plan, "num_points", 0):
+            self.maxPointsDetected.emit(len(points))
```
fdrgsp, line 144: The `QGraphicScene` coordinate are inverted compared to the one we use in `useq` so maybe it is worth adding this here

```suggestion
            self.maxPointsDetected.emit(len(points))

        # invert the y axis just for visualization purposes since point are calculated
        # in a coordinate system where the y axis is +up -down. In QGraphicScene the
        # y axis is +down -up.
        points = [p.replace(x=p.x, y=-p.y) for p in points]
        
```
We go form this:
Screenshot 2024-07-09 at 9 52 47 AM
to this:
Screenshot 2024-07-09 at 12 11 01 PM

tlambert03: good catch.  
btw: in general we use the `.replace` method that I added in useq schema too much.  it was meant to be a public convenience for mutating otherwise "frozen" items, with a) a slightly nicer api than the already-provided `model_copy` function that pydantic provides  and b) ensured validation (it actually dumps and recreates the full object).  However, if you *know* that the value you are replacing it with is valid, then you shouldn't use `.replace`, you should use `model_copy` 👍

tlambert03:
ok, I think all the things I wanted to deal with (besides tests) are now dealt with.  Take another go at using it, and let me know how it feels.  then I'll bring over the tests you wrote for this part, and fix them to work with any changes

src/pymmcore_widgets/hcs/_fov_widget/_fov_widget.py
```diff
@@ -0,0 +1,58 @@
...
+        self.valueChanged.emit(value)
+
+    def _on_view_max_points_detected(self, value: int) -> None:
+        self.selector.random_points_wdg.num_points.setValue(value)
+
+    def _on_view_position_clicked(
+        self, index: int, position: useq.RelativePosition
+    ) -> None:
+        if self.selector.active_plan_type is useq.RandomPoints:
+            self.selector.random_points_wdg.start_at = position.replace(name=None)
```
fdrgsp, line 58: since we inverted the y axis, we should invert it here as well or we do not select the FOV we click on
```suggestion
            position = position.model_copy(update={"y":-position.y})
            self.selector.random_points_wdg.start_at = position.replace(name=None)
```

tlambert03: that's a bit confusing, I think instead we should store the unswapped position in the item.data()

tlambert03: fixed in 06846ba

tlambert03:
ok, tests added.  Assuming they pass, feel free to merge this when ready @fdrgsp, and add the further improvements in a followup

PR #318, opened 2024-07-09
feat: add useq.WellPlanPlan widget with well selection

tlambert03:
minimal working example of a plate selector extracted from @fdrgsp's code in #308 

Screenshot 2024-07-11 at 4 59 25 PM

:joy:

TODO:

- [x] add back well names
- [x] support selection

PR #319, opened 2024-07-10
feat: add restrict well area

fdrgsp:
- restrict well area
- test


src/pymmcore_widgets/useq_widgets/points_plans/_well_graphics_view.py
```diff
@@ -70,40 +72,85 @@ def setPointsPlan(self, plan: useq.RelativeMultiPointPlan) -> None:
...
+            QRectF(-width / 2, -height / 2, width, height)
+            if self._is_circular
+            else QRectF(-(width - (width / 2)), -(height - (height / 2)), width, height)
+        )
+
+        pen = QPen(QColor(Qt.GlobalColor.green))
+        pen.setWidth(self._scaled_pen_size())
+        pen.setStyle(Qt.PenStyle.DotLine)
+
+        if self._is_circular:
```
tlambert03, line 121: technically the bounding area should come from the `RandomPoints.shape`, where `self._is_circular` should come from `WellPlate.circular_wells`.  i.e. there's nothing conceptually wrong with constraining the points into a square bounding area inside a circular well

src/pymmcore_widgets/useq_widgets/points_plans/_well_graphics_view.py
```diff
@@ -70,40 +72,85 @@ def setPointsPlan(self, plan: useq.RelativeMultiPointPlan) -> None:
...
 
+    def _draw_fov_constrain_area(self, width: float, height: float) -> None:
+        """Draw the bounding area to constrain the FOVs."""
+        if self._fov_constrain_area_item:
+            self._scene.removeItem(self._fov_constrain_area_item)
+
+        if self._well_width_um is None or self._well_height_um is None:
+            return
+
+        # bounding area to constrain the FOVs
```
tlambert03, line 110: can you follow the same pattern above with 

```python
        if (rect := self._points_bound_rect()).isNull():
             return
```

src/pymmcore_widgets/useq_widgets/points_plans/_points_plan_widget.py
```diff
@@ -52,15 +76,49 @@ def __init__(
         self._well_view.maxPointsDetected.connect(self._on_view_max_points_detected)
         self._well_view.positionClicked.connect(self._on_view_position_clicked)
 
-        if plan is not None:
-            self.setValue(plan)
+        # if plan is not None:
+        self.setValue(plan or INIT_PLAN)
+
+        # init the view with the current well size
+        self.setWellSize(self._well_width_µm or 6, self._well_height_µm or 6)
```
tlambert03: isn't `_well_width_µm` in microns and `6` in mm?

fdrgsp: yes 👍🏻, I changed that in [1c03208](https://github.com/pymmcore-plus/pymmcore-widgets/pull/319/commits/1c0320808f928c183296071f89d59f5898bd2aa3)

src/pymmcore_widgets/useq_widgets/points_plans/_well_graphics_view.py
fdrgsp: so do you want to add a `well_plate: WellPlate | None = None` arg to the `PointsPlanWidget`?

fdrgsp: or maybe something more general like `well_info: useq.WellPlate | tuple[float, float] | float | None = None` which can include either a `WellPlate` or simply a `well_size`?

tlambert03: i don't think it needs awareness, or a pointer to, a full plate object.  It just needs to track and retain the difference in meaning between the shape of the "world" in which this single well lives (well size and shape) and the "bounds" in which the points live (bounds size and shape).  when used in the *context* of some larger widget that manages plates and stuff, that outer widget might choose to set the well size/shape of this widget accordingly, but this widget itself doesn't need to know anything about a full well plate

fdrgsp:
@tlambert03 I'm done with the tests and are covering most of the widget. You can have a look if you want!

tlambert03:
not sure what's going on with that one 3.8 pyside2 test... not too worried,  it's failing in #318 as well

PR #327, opened 2024-07-13
fix: WellPlateWidget initial drawing

fdrgsp:
currently if we create the `WellPlateWidget` without passing a `plate` arg, the plate is not drawn even if there is a plate selected in the combobox. This will fix that.

Screenshot 2024-07-12 at 9 51 09 PM

Screenshot 2024-07-12 at 9 51 24 PM


tlambert03:
thanks, this is in #322 as well, let's use that implementation, which is a bit more direct:

https://github.com/pymmcore-plus/pymmcore-widgets/pull/322/files#diff-e61f82640f1bbe3e269fae5e5b541165750e64fb403b2ae4312cd7e96451c82aR117-R118

tlambert03:
scratch that... i like this one better

tlambert03 (approved):

PR #334, opened 2024-07-16
refactor: refactor stage widget

### LINKED ISSUE #331: Stage Control Widget Checkboxes appear disabled
* Python version: 3.12.4

### Description

I was showing the Stage Control widgets to a colleague, who remarked that he hadn't tried clicking the checkboxes at the bottom because they appeared disabled.

It would be great if the text was the same color as "Set as Default".

Also of note is the cut-off checkboxes, however that seems to be a Windows thing that is probably tricky to fix. I'm not sure if it's worth a separate issue.

### What I Did

Put the Stage Control widgets in my UI :

image

Alternatively, the behavior is simply seen in the [documentation](https://pymmcore-plus.github.io/pymmcore-widgets/widgets/StageWidget/) screenshot.

### LINKED ISSUE #332: Stage Control - consider absolute position field
* Python version: 3.12.4

### Description

A colleague noted that sometimes he knows exactly where he wants a stage to be positioned, and suggested it would be nice to enter the absolute stage position somewhere on the widget. One option could be to replace the circled `QLabel` with a couple of `QDoubleSpinBox`es?

image

tlambert03:
closes #331 

edit:   let's do #332 elsewhere

### DIFF
--- a/src/pymmcore_widgets/_stage_widget.py
+++ b/src/pymmcore_widgets/_stage_widget.py
@@ -1,14 +1,15 @@
 from __future__ import annotations
 
-from itertools import chain, product, repeat
-from typing import ClassVar
+from itertools import product
+from typing import cast
 
 from fonticon_mdi6 import MDI6
-from pymmcore_plus import CMMCorePlus, DeviceType
-from qtpy.QtCore import Qt, QTimer
+from pymmcore_plus import CMMCorePlus, DeviceType, Keyword
+from qtpy.QtCore import Qt, QTimerEvent, Signal
 from qtpy.QtWidgets import (
     QCheckBox,
     QDoubleSpinBox,
+    QFormLayout,
     QGridLayout,
     QHBoxLayout,
     QLabel,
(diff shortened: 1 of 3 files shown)


tlambert03:
i made the position labels editable, as per #332 ... however now that I'm going to hook them up to the core to actually move the stage, I'm realizing that's probably a bit unsafe?   It's much easier to enter an invalid number via text than it is to hit a move button too many times.

We should perhaps have a "halt" button or something?

src/pymmcore_widgets/_stage_widget.py
```diff
@@ -75,153 +190,141 @@ class StageWidget(QWidget):
...
-        self._update_position_label()
-
-        self._poll_cb = QCheckBox("poll")
-        self._poll_cb.setMaximumWidth(50)
-        self._poll_timer = QTimer()
-        self._poll_timer.setInterval(500)
-        self._poll_timer.timeout.connect(self._update_position_label)
-        self._poll_cb.toggled.connect(self._toggle_poll_timer)
+        self._poslabel_x = QDoubleSpinBox()
+        self._poslabel_x.setButtonSymbols(QSpinBox.ButtonSymbols.NoButtons)
```
gselzer, line 226: Similar to my above note, Do you think we could show buttons, not because we want people to use them, but because we want people to know that they can edit the value and the stage will react? Or do you have a better idea for conveying this idea?

src/pymmcore_widgets/_stage_widget.py
```diff
@@ -75,153 +190,141 @@ class StageWidget(QWidget):
...
+        main_layout.setContentsMargins(5, 5, 5, 5)
+        main_layout.addWidget(self._set_as_default_btn, 0, Qt.AlignmentFlag.AlignCenter)
+        main_layout.addWidget(self._move_btns, Qt.AlignmentFlag.AlignCenter)
+        main_layout.addLayout(chxbox_grid)
+
+        # position label
+        # this can appear either below or to the right of the move buttons
+
+        if position_labels_below is None:
+            self._pos_labels_below = is_2axis
```
gselzer, line 263: I think this appears a bit strange to me, because if you have an XY stage and a Z stage side-by-side, the labels will appear differently despite passing the same arguments to the widget. i.e. given [this example](https://pymmcore-plus.github.io/pymmcore-widgets/widgets/StageWidget/#example) you'll have something like this:

image

I'd prefer that if `position_labels_below` is `None`, then it should be `True`. Was there a reason you decided on this logic?

gselzer (commented):
Thanks for moving on this so quickly @tlambert03!

I didn't look too hard over the code, but I tried it within my UI and commented on the things I saw (I use these within a widget that is very similar to [this example](https://pymmcore-plus.github.io/pymmcore-widgets/widgets/StageWidget/#example))

Other things that came to mind:
* Is there any way for us to change the relative difference in step size? i.e. what if instead of `2 x stepSize` for the double chevron, you wanted 10x?
* What if, instead of always moving in microns, we could consider changing the units to something else, like "fields"? We could add a unit dropdown next to the step size, although I'm not sure how it'd look. I'm also not sure whether it is within the scope of this PR, but it is related...

tlambert03:
> Is there any way for us to change the relative difference in step size? i.e. what if instead of 2 x stepSize for the double chevron, you wanted 10x?

not at the moment, that would be a good feature.  The MMStudio one lets you specify the step for each individual arrow:

Screenshot 2024-07-16 at 6 13 27 PM

it's a nice feature but obviously takes up more space, making it hard to have this widget part of a dock widget layout by default.  so, in future PRs, lets consider ways to expose that functionality in a space efficient way.

> What if, instead of always moving in microns, we could consider changing the units to something else, like "fields"?

😂 did you or the suggester already use the MMStudio widget above?  or just come up with that on your own :)  yep, a way to auto-populate steps based on fields would be nice too.  again, let's do in a future PR

src/pymmcore_widgets/_stage_widget.py
```diff
@@ -21,37 +22,150 @@
...
+        btn_grid = QGridLayout(self)
+        btn_grid.setContentsMargins(0, 0, 0, 0)
+        btn_grid.setSpacing(0)
+        for glyph, (row, col, xmag, ymag) in MOVE_BUTTONS.items():
+            btn = MoveStageButton(glyph, xmag, ymag)
+            btn.clicked.connect(self._on_move_btn_clicked)
+            btn_grid.addWidget(btn, row, col, Qt.AlignmentFlag.AlignCenter)
+
+        # step size spinbox in the middle of the move buttons
+        self.step_size = QDoubleSpinBox()
```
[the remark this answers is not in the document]
tlambert03: > add a unit here

👍 

> Is it possible to add some sort of indication (spin buttons?) to indicate that this value can be edited

yeah, it's possible.  again, i rather dislike the look/UX of spin buttons; and I do think that the fact that the cursor turns into the little text cursor icon when you hover over it goes a long way towards making you realize that.  One is very likely to accidentally brush over the field while using the widget, and then they know forever that it's editable.  so, would love to avoid adding the spin buttons if at all possible.  if the hover effects aren't enough, let's brainstorm on other ways to make it obvious without spin buttons

src/pymmcore_widgets/_stage_widget.py
tlambert03: same note as above :) dislike the spinbuttons, like the hover effects, think it's quickly learnable.  spin buttons as a last resort, would prefer a different option

src/pymmcore_widgets/_stage_widget.py
tlambert03: i reverted that change.

Note that the example is also being [changed in this PR](https://github.com/pymmcore-plus/pymmcore-widgets/pull/334/files#diff-b3b7d3f2b3c3ba8a4f75f2be59d2bda2eaf7da47e63d27ecba02b133c9c2deecR20-R27), so had you followed the example after this PR (which encourages being explicit about your desired layout if you're using them in a group), it wouldn't look like that.  but it's fine with me to fallback to always having a vertical layout

src/pymmcore_widgets/_stage_widget.py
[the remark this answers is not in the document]
tlambert03: btw, can't show it in a screenshot, but this is what the cursor looks like when you hover past that field:

cursor-ibeam

src/pymmcore_widgets/_stage_widget.py
```diff
@@ -75,153 +204,144 @@ class StageWidget(QWidget):
...
-        self.layout().addWidget(top_row)
-        self.layout().addWidget(self._btns, AlignCenter)
-        self.layout().addWidget(bottom_row_1)
-        self.layout().addWidget(bottom_row_2)
-
-        if self._dtype is not DeviceType.XYStage:
+        self._invert_y = QCheckBox(text=f"Invert {Y}")
+        self._set_as_default_btn = QRadioButton(text="Set as Default")
+        # no need to show the "set as default" button if there is only one device
+        if len(self._mmc.getLoadedDevicesOfType(self._dtype)) < 2:
```
gselzer, line 256: I was actually about to file an issue for this, you're too fast! 💨

PR #338, opened 2024-07-16
[WIP] feat: new widget to manage Groups and Presets

fdrgsp:
@tlambert03 I started from your nicel #307 PR and I've use the logic to create a new `GroupPresetWidget`. I believe it is easier to manage the Micro-Manager Groups and Presets (when compared to our current widget). I'm sure we can improve this but I think it is a good starting point.

https://github.com/user-attachments/assets/ded6229e-9923-4197-908d-e2666eb60537

TODO:

- [ ] add test


tlambert03:
I'm going to close this since i think #307 will encompass all the use cases here

PR #343, opened 2024-07-17
style: Manually compute sizeHint()

gselzer:
Thanks to @tlambert03 for the suggestion!

Here's what it looks like on Windows 11, and it also looks good on WSL - can someone else confirm this looks good to them?

image

### DIFF
--- a/src/pymmcore_widgets/_pixel_configuration_widget.py
+++ b/src/pymmcore_widgets/_pixel_configuration_widget.py
@@ -473,9 +473,7 @@ class AffineTable(QTableWidget):
 
         self.horizontalHeader().setSectionResizeMode(QHeaderView.ResizeMode.Stretch)
         self.horizontalHeader().setVisible(False)
-        self.verticalHeader().setSectionResizeMode(
-            QHeaderView.ResizeMode.ResizeToContents
-        )
+        self.verticalHeader().setDefaultSectionSize(20)
         self.verticalHeader().setVisible(False)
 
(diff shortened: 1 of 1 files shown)


src/pymmcore_widgets/_pixel_configuration_widget.py
```diff
@@ -473,7 +473,9 @@ def __init__(self, parent: QWidget | None = None):
 
         self.horizontalHeader().setSectionResizeMode(QHeaderView.ResizeMode.Stretch)
         self.horizontalHeader().setVisible(False)
-        self.verticalHeader().setSectionResizeMode(QHeaderView.ResizeMode.Stretch)
+        self.verticalHeader().setSectionResizeMode(
+            QHeaderView.ResizeMode.ResizeToContents
+        )
```
tlambert03, line 478: what happens if you use 

```suggestion
        self.verticalHeader().setDefaultSectionSize(20)
```

here instead?

I know it feels like a magic number (and it is; albeit a very common qt line height) ... however with resizeToContents... I see this (extra room at the bottom)

Screenshot 2024-07-17 at 2 33 34 PM

and with an explicit section size I see:

Screenshot 2024-07-17 at 2 33 21 PM

what does explicit section size look like for you?

gselzer: It looks fine to me - switched it out in 802fe53. It must have been the `ResizeMode.Stretch` that was causing the problem.

Issue #339, opened 2024-07-17
All Radio Buttons appear On within GridPlan widget on Windows 11


gselzer:
* pymmcore-widgets version: 0.7.2
* Python version: 3.12.4
* Operating System: Windows 11

### Description
A colleague was confused by the radio buttons within the `GridPlanWidget`, as they appeared to him to all be on at once - it's not apparent at a glance which is actually on.

![image](https://github.com/user-attachments/assets/d9b2c49a-a7df-4d2c-b396-122cf3284dc3)

The UI does not have this problem on macOS (at least as far as I can tell on the documentation screenshots) nor on WSL, so it might be a OS-specific thing we can't do much about, or maybe it is a styling thing?

tlambert03:
this is what I see on windows 10:

![Capture](https://github.com/user-attachments/assets/5271ec62-0bbd-46fc-a0ea-2365a17982c1)

what does yours like like without the dark theme?

gselzer:
Nope 😅

![image](https://github.com/user-attachments/assets/385061a3-b73f-47b2-9073-4f7dd2a22b85)

Windows 11 thing, then?

tlambert03:
ok,  yeah, i think it's the accent color on windows 11 somehow...  try playing with your personalization stuff and see if you can learn anything.  That circle in the first one *is* showing that it's selected...

<img width="34" alt="image" src="https://github.com/user-attachments/assets/62d653af-5bf3-4342-8fe4-9ba16f121974">

but it's obviously incredibly hard to see.  windows is annoying.  I also don't know why your other fields don't seem to be disabled.   this is the standard example right?  Not a widget within a broader app that might have a style sheet applied?

gselzer:
> try playing with your personalization stuff and see if you can learn anything. 

Yeah, I'll keep looking.

> windows is annoying. 

💯

> I also don't know why your other fields don't seem to be disabled like they are in my view. 

Well, they *are* disabled, but there's no visual indication (outside of the cursor not changing to the I-Beam on the disabled fields) ☹️ 

> this is the standard example right? Not a widget within a broader app that might have a style sheet applied?

The above image was from my UI, however it looks identical if I run `python pymmcore-widgets/examples/mda_widget.py`

gselzer:
So this MCVE showcases the problem. On Windows 11, it would seem that the line `self.setWidget(self.inner)`  is the culprit that affects the style, as removing that line removes the accent fill that makes this so confusing.

[...]

PR #344, opened 2024-07-18
style: unfill radio buttions in GridPlanWidget

gselzer:
This PR, closing #339, adds a `HACK` to prevent the OS accent color from filling up the `GridPlanWidget`'s radio buttons on Windows 11. This `HACK` seems to be the least-bad of the different options that could resolve the issue, which are:
* Setting the style sheet, as I did here, to override the accent color fill
* Adding the radio buttons to the layout of the `QScrollArea` itself, which is worse as it alters the hierarchy of the widget.

@tlambert03 and/or @fdrgsp: I'd be happy for your opinions and/or testing. It looks nice to me on Windows 11, and on WSL (Wayland)


gselzer:
Closes #339

tlambert03 (approved):
hilarious :joy: ... works for me!  looks exactly the same as before on macos

PR #347, opened 2024-07-18
fix: Only allow YAML save/load when YAML available

### LINKED ISSUE #333: Saving MDASequence as YAML crashes if PyYAML not installed
* Python version: 3.12.4

### Description

By default, PyYAML is not installed (it's an [extension of `useq-schema`](https://github.com/pymmcore-plus/useq-schema/blob/3eea067cfbc922bcd602c18d164c4be446f0d062/pyproject.toml#L40)). If you try to [save a MDASequence](https://github.com/pymmcore-plus/pymmcore-widgets/blob/5957860901237424a297096b4817e80f30927605/src/pymmcore_widgets/useq_widgets/_mda_sequence.py#L451) as YAML, it'll crash. 

Ideally, we'd either always depend upon PyYAML or we'd disallow saving as YAML if it is not installed.

gselzer:
This PR avoids exceptions arising from the attempt to save/load MDA settings from YAML files, when pyyaml is not installed.

Notably, I made the decision to prevent the selection of a YAML file within the `QFileDialog`. We could have alternatively shown it, but prevented the exception (through a `try`/`except` or other means) instead. I prefer this method as it leaves no room for unexpected behavior (namely a user trying to save into a YAML file and then save not being completed).

Closes #333

### DIFF
--- a/src/pymmcore_widgets/useq_widgets/_mda_sequence.py
+++ b/src/pymmcore_widgets/useq_widgets/_mda_sequence.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from importlib.util import find_spec
 from itertools import permutations
 from pathlib import Path
 from typing import cast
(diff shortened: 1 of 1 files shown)


tlambert03 (approved):
very nice

gselzer:
Thanks codecov!

tlambert03:
lol

PR #350, opened 2024-07-19
refactor: split run mda in mda widget

wl-stepp:
Just a first thought potentially for viewer integration etc. Would give a little more fine-grained control when customizing/subclassing the MDAWidget.

### DIFF
--- a/src/pymmcore_widgets/mda/_core_mda.py
+++ b/src/pymmcore_widgets/mda/_core_mda.py
@@ -184,7 +184,17 @@ class MDAWidget(MDASequenceWidget):
         return get_next_available_path(requested_path=requested_path)
 
     def prepare_mda(self) -> bool | str | None:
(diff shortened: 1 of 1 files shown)


src/pymmcore_widgets/mda/_core_mda.py
```diff
@@ -183,8 +183,8 @@ def get_next_available_path(self, requested_path: Path) -> Path:
         """
         return get_next_available_path(requested_path=requested_path)
 
-    def run_mda(self) -> None:
-        """Run the MDA sequence experiment."""
+    def prepare_mda(self) -> bool | str | None:
+        """Prepare the MDA sequence experiment."""
```
tlambert03, line 187: looks good to me.  can we just update this docstring to explain how one should interpret the return type?

tlambert03 (approved):
🚀

Issue #342, opened 2024-07-17
GridPlanWidget's Overlap, Order, Relative To confusions


gselzer:
* pymmcore-widgets version: 0.7.2
* Python version: 3.12.4
* Operating System: Windows 11

### Description
A colleague was unaware that the `GridPlanWidget` options `Overlap`, `Order`, and `Relative To` pertained to all of the different mechanisms for specifying a `Grid Plan`. He did not even notice the `Relative To` dropdown, because it was hidden:

![image](https://github.com/user-attachments/assets/d2ff63e2-0a32-4d11-b5b7-dd0163026929)

One fix could be ensuring that the widget is tall enough such that all tabs fit without the need for a scrollbar. Beyond that, maybe we could organize the layout differently such that it's clearer these options pertain to all methods of grid specification? How might we better do that? We could copy these options into each section of the `GridPlanWidget`, or add a box around the three radio button options to make it clearer that the options below are not an "additional choice"?

tlambert03:
could combine the top three into a QTabWidget, that would have the advantage of only showing the active settings, while always showing the general options below.   The downside is that tabs really don't imply "selection".  
Alternatively, could use a QStackedWidget, with selection buttons at the top determining which of the widgets is shown/active.  I think that might be done in the Z tab (which itself needs some work).

there's definitely many ways this widget could be improved.  Someone just needs to play around with it a bit

> We could copy these options into each section of the GridPlanWidget

that would be too much space i think

> or add a box around the three radio button options to make it clearer that the options below are not an "additional choice"?

open to that as well

> ensuring that the widget is tall enough such that all tabs fit without the need for a scrollbar

that is too screen/user specific to be a good solution.  for anything more than a couple hundred pixels, you pretty much always have to allow for the possibility that it won't fit wherever it's being placed

gselzer:
This is solved now with #351

PR #351, opened 2024-07-19
feat: Refactor GridPlanWidget

gselzer:
This PR provides a different approach to constructing a Grid for an MDA sequence by refactoring the `GridPlanWidget`. It's a draft PR because the code is messy, and because I want feedback. It's perfectly fine if others prefer the existing `GridPlanWidget` and we discard these changes.

From UX testing with a collaborator, we filed issues such as #342 and #339, and the general impression that I got was that the `GridPlanWidget` could be slimmed down. I was thinking about the following goals when I constructed this new design:
* Shorten the `GridPlanWidget` so that it is more likely the full widget will fit within the `MDAWidget`, without a scroll bar.
* Shorten the distance between the "bottom stuff" (`Order`, `Overlap`, and `Relative To`) and the grid construction parameters being used. We heard that it might be difficult to associate the bottom stuff with the parameter sets towards the top, especially if some of them aren't seen (immediately) due to the first point.
* Avoid showing parameters that aren't currently being used. This applies both to parameter groups that aren't being used and to the bottom stuff (i.e. `RelativeTo`). I believe that this will result in a less overwhelming widget.
* Elaborate on the meaning of different parameters when I was unclear on what their purpose was.

The result, based on a `QStackWidget` instead of a `QButtonGroup`, looks like the following:

https://github.com/user-attachments/assets/91139c19-04a1-4443-88d6-27e5afb5440e

Pros:
* In my opinion, this widget better realizes each of the goals I listed above.
* We avoid the need for the HACK that I added in #344

Cons:
* Because the bottom stuff now is a part of each widget in the `QStackWidget`, we need multiple instances of the `QEnumComboBox`/`QDoubleSpinBox`, which need coordination. I added coordinating glue, but it's extra code to maintain.
* The result looks "more boring" in my opinion. Maybe it's a good thing, maybe not.

I'm looking for opinions from @tlambert03, @fdrgsp, @marktsuchida and others on whether this is an improvement, and what might make it better if we decide it should replace the original widget. If we do like it, I'm happy to make additional changes and further clean up the code.

### DIFF
--- a/src/pymmcore_widgets/useq_widgets/_grid.py
+++ b/src/pymmcore_widgets/useq_widgets/_grid.py
@@ -1,21 +1,18 @@
 from __future__ import annotations
 
 from enum import Enum
-from typing import Literal, Sequence, cast
+from typing import Literal
 
 import useq
 from qtpy.QtCore import QSize, Qt, Signal
 from qtpy.QtWidgets import (
-    QAbstractButton,
-    QButtonGroup,
     QDoubleSpinBox,
     QFormLayout,
     QGridLayout,
-    QHBoxLayout,
     QLabel,
-    QRadioButton,
     QScrollArea,
     QSpinBox,
+    QStackedWidget,
     QVBoxLayout,
     QWidget,
 )
@@ -41,9 +38,12 @@ class OrderMode(Enum):
 
 
 class Mode(Enum):
-    NUMBER = "number"
-    AREA = "area"
-    BOUNDS = "bounds"
+    NUMBER = "Fields of View"
+    AREA = "Width & Height"
+    BOUNDS = "Absolute Bounds"
+
+    def __str__(self) -> str:
+        return self.value
 
 
 class GridPlanWidget(QScrollArea):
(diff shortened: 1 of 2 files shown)


gselzer:
>     * why do the widgets inside the stack widget have to duplicate all the bottom stuff?  Why not just put only the stuff that changes inside the stack widget and leave the bottom stuff permanently visible?  Should require very little change?

I actually tried leaving them as a separate widget at first, however the `StackWidget` maintained constant size, meaning that the bottom stuff was vertically positioned to be below the bounds option. Then, when you instead had one of the other options selected, there was an unnatural gap between the option-specific parameters and the bottom stuff.

On a very minor point, I also like that the separator between the option-specific parameters and the bottom stuff is shorter (due to being within the same widget) than the one above - I think it better conveys the idea that the parameters belong to *this specific option*.

It may actually be possible to resize the stack widget based upon the current selection - I could investigate if you'd like.

>     * curious what three radio buttons in a buttongroup at the top would look like.  I find combo boxes slightly annoying when there aren't many options: radioboxes allow you to see what's available (provided they fit on one line) and change very quickly.

I could try it - granted, I'd have to add the Windows 11 radio button HACK back in 😅

tlambert03:
all of that visual stuff should be possible either way (both the sizing of the stack widget and the short separator line) regardless of how we structure the widgets... let me dig a bit and see what I can find.  it would make the code much easier to maintain, so would be preferable (though I definitely agree that appearance is important too)

tlambert03:
found this, but it didn't work immediately for me (everything was now too squashed): https://stackoverflow.com/questions/14480696/resize-qstackedwidget-to-the-page-which-is-opened

but it suggests it's possible...

would you mind taking a stab at another version of this branch? where:
1. create new widget classes for each mode:
      <details>
      
      <summary>something like ...</summary>
      
      ```python
      class _RowsColsWidget(QWidget):
          valueChanged = Signal()
      
          def __init__(self) -> None:
              super().__init__()
      
              self.rows = QSpinBox()
              self.rows.setRange(1, 1000)
              self.rows.setValue(1)
              self.rows.setSuffix(" fields")
              self.columns = QSpinBox()
              self.columns.setRange(1, 1000)
              self.columns.setValue(1)
              self.columns.setSuffix(" fields")
      
              layout = QFormLayout(self)
              layout.setFieldGrowthPolicy(QFormLayout.FieldGrowthPolicy.AllNonFixedFieldsGrow)
              layout.addRow("Rows:", self.rows)
              layout.addRow("Columns:", self.columns)
      
              self.rows.valueChanged.connect(self.valueChanged)
              self.columns.valueChanged.connect(self.valueChanged)
      
      
      class _WidthHeightWidget(QWidget):
          valueChanged = Signal()
      
          def __init__(self) -> None:
              super().__init__()
      
              self.area_width = QDoubleSpinBox()
              self.area_width.setRange(0.01, 100)
              self.area_width.setDecimals(2)
              # here for area_width and area_height we are using mm instead of µm because
              # (as in GridWidthHeight) because it is probably easier for a user to define
              # the area in mm
              self.area_width.setSuffix(" mm")
              self.area_width.setSingleStep(0.1)
              self.area_height = QDoubleSpinBox()
              self.area_height.setRange(0.01, 100)
              self.area_height.setDecimals(2)
              self.area_height.setSuffix(" mm")
              self.area_height.setSingleStep(0.1)
      
              layout = QFormLayout(self)
              layout.setFieldGrowthPolicy(QFormLayout.FieldGrowthPolicy.AllNonFixedFieldsGrow)
              layout.addRow("Width:", self.area_width)
              layout.addRow("Height:", self.area_height)
      
              self.area_width.valueChanged.connect(self.valueChanged)
              self.area_height.valueChanged.connect(self.valueChanged)
      
      
      class _BoundsWidget(QWidget):
          valueChanged = Signal()
      
          def __init__(self) -> None:
              super().__init__()
              self.left = QDoubleSpinBox()
              self.left.setRange(-10000, 10000)
              self.left.setValue(0)
              self.left.setDecimals(3)
              self.left.setButtonSymbols(QDoubleSpinBox.ButtonSymbols.NoButtons)
              self.top = QDoubleSpinBox()
              self.top.setRange(-10000, 10000)
              self.top.setValue(0)
              self.top.setDecimals(3)
              self.top.setButtonSymbols(QDoubleSpinBox.ButtonSymbols.NoButtons)
              self.right = QDoubleSpinBox()
              self.right.setRange(-10000, 10000)
              self.right.setValue(0)
              self.right.setDecimals(3)
              self.right.setButtonSymbols(QDoubleSpinBox.ButtonSymbols.NoButtons)
              self.bottom = QDoubleSpinBox()
              self.bottom.setRange(-10000, 10000)
              self.bottom.setValue(0)
              self.bottom.setDecimals(3)
              self.bottom.setButtonSymbols(QDoubleSpinBox.ButtonSymbols.NoButtons)
      
              layout = QFormLayout(self)
              layout.setFieldGrowthPolicy(QFormLayout.FieldGrowthPolicy.AllNonFixedFieldsGrow)
              layout.addRow("Left:", self.left)
              layout.addRow("Top:", self.top)
              layout.addRow("Right:", self.right)
              layout.addRow("Bottom:", self.bottom)
      
              self.top.valueChanged.connect(self.valueChanged)
              self.bottom.valueChanged.connect(self.valueChanged)
              self.left.valueChanged.connect(self.valueChanged)
              self.right.valueChanged.connect(self.valueChanged)
      
      ```
      </details>

1. put those in a stacked widget, perhaps playing with the pattern linked above.
1. give each of them some sort of a `values()` method that returns `dict[str, Any]` that can be merged into the other values inside of `GridPlanWidget.value()`, and similarly a `set_values`
1. deal with the changes to that modes thing (which will get a lot cleaner after we have individual widgets)

might not work perfectly at first, but i'd like to see that pattern

gselzer:
@tlambert03 here's the version with radio buttons. I like it better! What do you think?

image

fdrgsp:
This is nice, I like it too!

marktsuchida:
I like this, too; the radio buttons at the top are nice!

gselzer:
@tlambert03 was able to revert to the bottom stuff being in a separate widget, with essentially the same look (I did remove the separator above the bottom stuff, though). Let me know what you 
image
think!

src/pymmcore_widgets/useq_widgets/_grid.py
```diff
@@ -368,3 +269,166 @@ def _on_change(self) -> None:
...
+        self.overlap.setSuffix(" %")
+        self.order_lbl = QLabel("Acquisition order:")
+        self.order = QEnumComboBox(self, OrderMode)
+        self.relative_to_lbl = QLabel("Current position:")
+        self.relative_to = QEnumComboBox(self, RelativeTo)
+
+        self.form_layout = QFormLayout(self)
+
+        # NB relative_to added in self.setMode
+        self.form_layout.addRow(self.overlap_lbl, self.overlap)
```
tlambert03, line 418: if you want that short line here, which i also liked, add:

```suggestion
        self.form_layout.addRow("", SeparatorWidget())
```

tlambert03 (commented):
good stuff!  I'm happy to get this in as is... couple small ideas for follow ups, but they don't need to be in here.  if you can get the tests working again this can go in

gselzer:
@tlambert03 I addressed your comments and fixed the tests. I feel like a squash+merge is appropriate, but let me know if there's anything else I can do!

tlambert03:
thanks a bunch!  much nicer

PR #353, opened 2024-07-22
feat: reusable single-well calibration widget for plate calibration widget

### LINKED ISSUE #329: todo: plate calibration widget
this issue is just to discuss design strategies around the plate calibration widget in #308  (and #322)
It's a complicated widget and I think we need to build it up piecewise.

I think the first step is to make a reusable single-well calibration widget. It's goal is simply to find the center of a single well
it should have

1. a label at the top to display which well is currently being calibrated ("Well A1")
2. Some sort of green/yellow visual representation next to that label to show whether the well is calibrated or not
3. a method/mode combobox (similar to the one at the top in #308), where the options depend on circular/square
    1. Circular methods:
        - 1 center point
        - 3 edge points
    2. Square methods
        - 1 center point
        - 2 corners
        - 4 edge points
4. Below that is a table.  In all cases, the table should be *pre* populated with the number of points required, with the value as `-` if it hasn't been set yet.  In the square cases, it's possible (but not clear to me yet) that we need to specify exactly which of the 2/4 points the user should be entering.  (I know would could try to guess that from what the user enters, but we'd be making assumptions about plate orientation).  In which case the table needs to be three columns (point, X (mm), Y(mm))
5. Below that is a big "set" button, that, when clicked, populates the selected row in the table with the current stage position from core
6. Below that are two buttons to 1) clear the currently selected row 2) clear all rows

considerations:
- perhaps provide a well size and error threshold:  if we *know* we're calibrating a well with a given size, then the provided points should be the appropriate distance from the calibrated center point, within some threshold, or validation fails
 
by default, I actually think the table should not be manually editable (though during development 

here's a stub:

```python
from qtpy.QtCore import Signal
from qtpy.QtWidgets import QComboBox, QTableWidget, QTableWidgetItem, QWidget


class WellCalibrationTable(QWidget):
    # emitted when the required number of points have been selected and validated
    calibrationChanged = Signal(bool)

    def __init__(self) -> None:
        ...
        self._combo = QComboBox()
        self._combo.currentTextChanged.connect(self._set_mode)

        self._table = QTableWidget()
        self._table.itemChanged.connect(self._validate_table)
        self._table.setSelectionMode(QTableWidget.SelectionMode.SingleSelection)
        self._table.setSelectionBehavior(QTableWidget.SelectionBehavior.SelectRows)

    def setCircularWell(self, circular: bool) -> None:
        """Populate the combobox with either circular or rectangular options."""

    def _set_mode(self, mode: str) -> None:
        """Set the mode for the calibration table.

        - 1 center point (either)
        - 2 edge points (square)
        - 3 edge points (circle)
        - 4 edge points (square)
        """
        self._table.setRowCount(...)

    def _validate_table(self, item: QTableWidgetItem) -> None:
        """Validate points.

        - if the new item is already in the list, show an error message and remove it.
        - if the number of points is not yet satisfied, just do nothing
        - if the number of points is correct, try to calculate the calibration,
          if it fails, show an error message. If it succeeds, emit
          calibrationChanged(True)
        - if we somehow went from a valid state to an invalid state, emit
          calibrationChanged(False)
        """

    def wellCenter(self) -> tuple[float, float] | None:
        """If calibrated, return the calculated center of the well."""
```

fdrgsp:
Closes #329 

Screenshot 2024-07-21 at 10 01 49 PM

to run the example:
```bash
python examples/temp/well_calibration_widget.py
```

TODO:
- [x] add tests


tlambert03:
ok @fdrgsp, I made some changes (some kinda significant).  please play with the behavior and see if I've made anything worse :joy:

tlambert03:
it does look a bit different:

Screenshot 2024-07-23 at 3 49 25 PM

fdrgsp:
> ok @fdrgsp, I made some changes (some kinda significant). please play with the behavior and see if I've made anything worse 😂

Looks good,  the behavior is similar! Do you want to rename the `set` button to `add` when more than 1 point? and what do you think about centering the well name and icon?

tlambert03:
>  Do you want to rename the set button to add when more than 1 point?

I don't think so, since the behavior is always to "set" the currently selected row to the current core position.

> what do you think about centering the well name and icon?

I undid that since i can imagine wanting to use it not centered (for example. in the context of the calibration widget, if it's off to the right)... in general, I'd prefer the widget that *consumes* this widget to make stylistic decisions, rather than this widget

tlambert03:
side note, played around a bit with autogenerating these svgs: 

```python
import math
import tempfile
from typing import Sequence

from qtpy.QtGui import QIcon
from qtpy.QtWidgets import QApplication, QPushButton


def circle_points_svg(
    view_box: int = 24,
    circle_radius: float = 8,
    line_width: float = 2,
    line_color: str = "#000000",
    num_points: int = 3,
    point_radius: float = 3,
    circle_color: str = "#00C600",
) -> str:
    vb = view_box
    svg = f"""<svg viewBox="0 0 {vb} {vb}" xmlns="http://www.w3.org/2000/svg">"""
    # Draw the circle
    svg += f"""<circle
        style="fill:none;stroke:{line_color};stroke-width:{line_width}"
        cx="{vb / 2}" cy="{vb / 2}" r="{circle_radius}" />
    """

    # Draw the points
    if num_points == 1:
        # draw a single point in the center
        svg += f"""<circle
            style="fill:{circle_color}"
            cx="{vb / 2}" cy="{vb / 2}" r="{point_radius}" />
        """
    else:
        for i in range(num_points):
            x = circle_radius * math.cos(2 * math.pi * i / num_points) + vb / 2
            y = circle_radius * math.sin(2 * math.pi * i / num_points) + vb / 2
            svg += f"""<circle
                style="fill:{circle_color}"
                cx="{x}" cy="{y}" r="{point_radius}" />
            """

    # Close the svg
    svg += "</svg>"
    return svg


def square_points_svg(
    view_box: int = 24,
    line_width: float = 1.6,
    asp_ratio: float = 1.1,
    line_color: str = "#000000",
    point_radius: float = 2.3,
    square_color: str = "#00C600",
    num_points: int = 4,
    coords: Sequence[tuple[float, float]] | None = None,
) -> str:
    vb = view_box
    svg = f"""<svg viewBox="0 0 {vb} {vb}" xmlns="http://www.w3.org/2000/svg">"""
    # Draw the square
    height = vb / 2
    width = height * asp_ratio
    left = (vb - width) / 2
    top = (vb - height) / 2
    svg += f"""<rect
        style="fill:none;stroke:{line_color};stroke-width:{line_width}"
        x="{left}" y="{top}" width="{width}" height="{height}" />
    """

    # Draw the points
    if coords is None:
        if num_points == 1:
            coords = [(vb / 2, vb / 2)]
        elif num_points == 2:
            # draw on opposite corners
            coords = [(vb / 4, vb / 4), (3 * vb / 4, 3 * vb / 4)]
        else:
            # Draw in the middle of top/left/bottom/right, slightly offset
            offset = 1.5
            coords = [
                (offset + left + width / 2, top),  # top - middle
                (left, height + offset),  # left - middle
                (left - offset + width / 2, 3 * top),  # bottom - middle
                (left + width, height - offset),  # right - middle
            ]

    for x, y in coords:
        svg += f"""<circle
            style="fill:{square_color}"
            cx="{x}" cy="{y}" r="{point_radius}" />
        """

    # Close the svg
    svg += "</svg>"
    return svg


def temp_svg_file(svg: str) -> str:
    f = tempfile.NamedTemporaryFile(suffix=".svg", delete=False)
    f.write(svg.encode())
    f.flush()
    return f.name


if __name__ == "__main__":
    app = QApplication([])
    widget = QPushButton()
    widget.setIcon(QIcon(temp_svg_file(square_points_svg())))
    widget.setIconSize(widget.size())
    widget.show()
    app.exec()
```

since these may not look good on a dark theme, for example...

tlambert03:
(let's deal with the icons later.  it's fine to commit them here for now, and that code above can just live in this issue).

@fdrgsp, can you write tests enough to get coverage passing?

fdrgsp:
@tlambert03 I've added fe tests 👍🏻

tlambert03:
> I've added fe tests

fank you

PR #355, opened 2024-07-24
feat: plate calibration widget

tlambert03:
Screenshot 2024-08-03 at 2 40 14 PM

```
$ python examples/temp/plate_calibration_widget.py
```

todo:

- [x] calculate a1_origin and rotation from 2+ calibrated wells
- [x] give some feedback info on remaining wells
- [ ] allow knowledge of plate to constrain the well calibration (for example, rejecting a well calibration if it falls too far outside of the known radius/size of a well)
- [ ] ~implement testing logic, to go to a well (center/edge)~ (hid the widget instead)
- [x] add tests
- [x] double check units on the well calibration table


fdrgsp:
@tlambert03 I love it!!!

PR #356, opened 2024-07-25
feat: plate navigator for HCS calibration testing

fdrgsp:
@tlambert03 what do you think to use something like this for the calibration testing of the HCS wizard? Basically the user has the freedom to click anywhere in the well to move to the stage...

https://github.com/user-attachments/assets/4c76fe7f-7333-4a23-864e-4708331a94b7

- [x] fix rotation
- [x] add tests


tlambert03:
love it. we certainly need something like that.  When you click does it always go to the center?  something that some commercial software add is the ability to go to top/left/bot/right/center:

image

(or, in the corners for a square well).  If you're trying to calibrate a plate, those would likely be more useful places to visit?  of course, as the number of wells goes up, it would be hard to show all of those points

fdrgsp:
> When you click does it always go to the center? 

At the moment, the stage goes wherever you click in the well (not always in the center). But we can make fixed position per well I guess!

tlambert03:
ah!  i see, yeah that's cool too.  i can imagine two different modes i guess.  one where you likely want specific easily identifiable places,  and the current one with complete freedom

fdrgsp:
> ah! i see, yeah that's cool too. i can imagine two different modes i guess. one where you likely want specific easily identifiable places, and the current one with complete freedom

hey @tlambert03 , what do you think about this:


https://github.com/user-attachments/assets/7c43bdde-2d12-4e81-8f3f-564f225faea0

tlambert03:
woah that's gorgeous

fdrgsp:
@tlambert03 I think this widget is functioning as we want, when you have time can you have a look at the code?

tlambert03 (approved):

PR #360, opened 2024-09-12
feat: add new calibration wdg

fdrgsp:
continue from #322


src/pymmcore_widgets/hcs/_hcs_wizard.py
```diff
@@ -0,0 +1,241 @@
...
+            path, _ = QFileDialog.getSaveFileName(
+                self,
+                "Save Well Plate Plan",
+                "",
+                "JSON or YAML (*.json *.yaml *.yml)",
+            )
+        if path:
+            value = self.value()
+            if value is None:
+                return
```
tlambert03, line 115: this is somewhat similar to the concept of disabling the Finish button if the plate is uncalibrated.  If you don't want the save button to do anything when the plate is not yet calibrated, then that button *must* be disabled until it's valid to click it.   It's not ok to let them click on it, pick a path, and then have it just close silently without doing anything

src/pymmcore_widgets/hcs/_hcs_wizard.py
```diff
@@ -0,0 +1,241 @@
...
+            if str(path).endswith((".yaml", ".yml")):
+                txt = value.yaml(exclude_unset=True)
```
tlambert03, line 117: didn't we run into this before? where pyyaml needs to be installed?  They shouldn't be able to hit this ImportError accidentally.  If you want to offer yaml as an export method, then you should first check whether yaml is available before allowing it.  Alternatively, just only accept json

src/pymmcore_widgets/hcs/_hcs_wizard.py
```diff
@@ -0,0 +1,241 @@
...
+        """Load a well plate plan from disk."""
+        if not isinstance(path, str):
+            path, _ = QFileDialog.getOpenFileName(
+                self,
+                "Load Well Plate Plan",
+                "",
+                "JSON or YAML (*.json *.yaml *.yml)",
+            )
+        if path:
+            self.setValue(WellPlatePlan.from_file(path))
```
tlambert03, line 132: same here. I think yaml is only allowed if pyyaml is installed.  we either need to commit to depending on pyyaml here, leave this sort of detail to the "application" that incorporates this widget, or include the conditional fallback here if pyyamls isnt' installed

tlambert03:
i broke a test here in [96a7820](https://github.com/pymmcore-plus/pymmcore-widgets/pull/360/commits/96a78200efc107a21626da7b434f7e0d0e19bfee) ... will have a look soon

tlambert03:
thanks for the improvements @fdrgsp.  I just added another simply test that should pass (basically, setting the wizard with a value and then retrieving that value without doing anything and checking that it's the same).  But it's failing at the moment, because it looks like origin information is lost.  Can you have a look and fix when you have a moment?

tlambert03:
(tests fixed in [d591308](https://github.com/pymmcore-plus/pymmcore-widgets/pull/360/commits/d59130819eef745f578ae8ad00a1666e06c23b73))

tlambert03:
going to merge this @fdrgsp 👍. great work!

PR #362, opened 2024-10-01
feat: add HCSWizard to MDAWIdget

fdrgsp:
this PR adds the `HCSWizard` to the `MDAWidget`.

- [ ] docs

Screenshot 2024-10-01 at 9 49 50 AM

Screenshot 2024-10-02 at 10 26 56 AM


src/pymmcore_widgets/mda/_core_mda.py
```diff
@@ -299,6 +330,76 @@ def _enable_widgets(self, enable: bool) -> None:
...
+            self._enable_position_table(False)
+            return
+        self._use_hcs = False
+        self._enable_position_table(True)
+
+    def _enable_positions_editing(self) -> None:
+        dialog = QMessageBox(
+            QMessageBox.Icon.Warning,
+            "Reset HCS",
+            "If you modify the positions, they will be considered as manually added, "
```
tlambert03, line 351: start with "Positions are currently autogenerated from the HCS Wizard.  Would you like to cast them to a list of stage positions?  Note: you will no longer be able to edit them using the HCS Wizard dialog."

- And then also, add a warning dialog when clicking on "add positions": "This will replace the positions currently stored in the table.  Would you like to proceed?"

- ideally: if the user clicks on the HCS wizard again after the positions are currently being controlled by the wizard, the wizard button should say "update positions"

tlambert03 (commented):
love it

fdrgsp:
hey @tlambert03, I think you can have another look!

tlambert03:
everything is looking good @fdrgsp, but i noticed that when you add positions from hcs (or go back to making them editable), the widget emits a lot of change events (i think it's around the number of positions in the well plate).  I added a failing test that demonstrates this.  Can you look into it?

fdrgsp:
> everything is looking good @fdrgsp, but i noticed that when you add positions from hcs (or go back to making them editable), the widget emits a lot of change events (i think it's around the number of positions in the well plate). I added a failing test that demonstrates this. Can you look into it?

@tlambert03 it should be fixed now.

tlambert03 (approved):
LGTM!

PR #378, opened 2024-11-05
fix: fix z plan and hardware autofocus plan validation error

### LINKED ISSUE #373: Bug in MDA widget
just hit this bug when clicking a bunch of buttons in the MDA widget (I think i clicked one of the hardware autofocus buttons last):

```pytb
  File "/Users/talley/dev/self/pymmcore-widgets/src/pymmcore_widgets/useq_widgets/_mda_sequence.py", line 524, in _update_time_estimate
    val = self.value()
          ^^^^^^^^^^^^
  File "/Users/talley/dev/self/pymmcore-widgets/src/pymmcore_widgets/mda/_core_mda.py", line 136, in value
    val = super().value()
          ^^^^^^^^^^^^^^^
  File "/Users/talley/dev/self/pymmcore-widgets/src/pymmcore_widgets/useq_widgets/_mda_sequence.py", line 407, in value
    val = val.replace(**replace)
          ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/talley/miniforge3/envs/pymmcore-widgets/lib/python3.12/site-packages/useq/_base_model.py", line 60, in replace
    return type(self).model_validate({**d, **kwargs})
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/talley/miniforge3/envs/pymmcore-widgets/lib/python3.12/site-packages/pydantic/main.py", line 596, in model_validate
    return cls.__pydantic_validator__.validate_python(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pydantic_core._pydantic_core.ValidationError: 1 validation error for MDASequence
  Value error, Absolute Z positions cannot be used with autofocus plan. [type=value_error, input_value={'metadata': {'pymmcore_w...hutter_open_across': ()}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.9/v/value_error
zsh: abort      python examples/mda_widget.py

```

fdrgsp:
tests failing because of https://github.com/hgrecco/pint/issues/2074.

fix the bug in #373.

@tlambert03 the error in #373 is due to the validation of the `useq.schema` [here](https://github.com/pymmcore-plus/useq-schema/blob/6dc031498f62c0d8853ada40188567f6d542335c/src/useq/_mda_sequence.py#L395-L402).  Basically we cannot use the hardware autofocus plan if an absolute z plan (top-bottom) is selected.

to fix that, I've created a method that disables the autofocus checkboxes (and thus remove the autofocus plan to avoid the validation error) if an absolute z plan is selected. If any other z plans (or None) are selected, the autofocus checkboxes are re-enabled. Do you agree with this solution?

closes #373

### DIFF
--- a/src/pymmcore_widgets/mda/_core_mda.py
+++ b/src/pymmcore_widgets/mda/_core_mda.py
@@ -339,6 +339,13 @@ class MDAWidget(MDASequenceWidget):
             self._mmc.mda.events.sequenceStarted.disconnect(self._on_mda_started)
             self._mmc.mda.events.sequenceFinished.disconnect(self._on_mda_finished)
 
+    def _enable_af(self, state: bool, tooltip1: str, tooltip2: str) -> None:
+        """Override the autofocus enablement to account for the autofocus device."""
+        if not self._mmc.getAutoFocusDevice():
+            self.stage_positions._update_autofocus_enablement()
+            return
+        return super()._enable_af(state, tooltip1, tooltip2)
+
 
 class _MDAControlButtons(QWidget):
     """Run, pause, and cancel buttons at the bottom of the MDA Widget."""
--- a/src/pymmcore_widgets/useq_widgets/_mda_sequence.py
+++ b/src/pymmcore_widgets/useq_widgets/_mda_sequence.py
@@ -500,6 +500,14 @@ class MDASequenceWidget(QWidget):
         # Only JSON
         return "All (*.json);;JSON (*.json)"
 
+    def _enable_af(self, state: bool, tooltip1: str, tooltip2: str) -> None:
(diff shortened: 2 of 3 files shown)


tlambert03:
looks like a step in the right direction, but I can still crash it by using z-top-bottom then clicking set AF offset per position in the position table 

Screenshot 2024-11-06 at 7 05 07 PM

tlambert03:
i wonder if there might be a more robust fix by also making sure that the method `MDASequenceWidget.value` itself can never return an invalid value

tlambert03 (approved):

PR #381, opened 2024-11-06
fix: fix include_z when setting stage positions value in PositionTable widget

### LINKED ISSUE #380: Bug: MDASequence useq widget does not load "include Z" option
* pymmcore-widgets version: pymmcore-widgets              0.1.dev248+g53e9ddf
* Python version: 3.11.9
* Operating System: Win10

### Description

When saving and reloading a MDASequence, the option to ignore Z will not be restored. Instead Z will always be there.

### What I Did

Slightly adjusted test_mda_wdg_load_save
```
import useq
from pymmcore_widgets.useq_widgets._mda_sequence import QFileDialog
from pymmcore_widgets.useq_widgets import MDASequenceWidget
import pymmcore_widgets
from pathlib import Path


MDA_noZ = useq.MDASequence(
    time_plan=useq.TIntervalLoops(interval=4, loops=3),
    stage_positions=[(0, 1), useq.Position(x=42, y=0)],
    channels=[{"config": "DAPI", "exposure": 42}],
    z_plan=useq.ZRangeAround(range=10, step=0.3),
    grid_plan=useq.GridRowsColumns(rows=10, columns=3),
    axis_order="tpgzc",
    keep_shutter_open_across=("z",),
)
ext= "json"
PYMMCW_METADATA_KEY = "pymmcore_widgets"

app = QApplication([])

wdg = MDASequenceWidget()
wdg.show()
import os
dest = Path(os.path.expanduser(f"~\Desktop\sequence.{ext}"))

dest.write_text("")

dest.write_text(MDA_noZ.yaml() if ext == "yaml" else MDA_noZ.model_dump_json())

wdg.load()
print(wdg.value().replace(metadata={}))
assert wdg.value().replace(metadata={}) == MDA_noZ

wdg.save()
mda_no_meta = MDA_noZ.replace(
    metadata={PYMMCW_METADATA_KEY: {"version": pymmcore_widgets.__version__}}
)

assert dest.read_text() == mda_no_meta.model_dump_json(exclude_defaults=True)

app._exec()
```

tlambert03:
fixes #380 

for now, this will set the "include z" checkbox to True if _any_ of the positions have a Z position.   If only some of them have a z position, it will emit a warning.

If we need the feature in the future, we could further elaborate to somehow let some rows have a Z position of None while others don't

### DIFF
--- a/src/pymmcore_widgets/mda/_core_positions.py
+++ b/src/pymmcore_widgets/mda/_core_positions.py
@@ -137,6 +137,7 @@ class CoreConnectedPositionTable(PositionTable):
             self._set_position_table_editable(False)
             value = tuple(value)
         super().setValue(value)
+        self._update_z_enablement()
 
     # ----------------------- private methods -----------------------
 
--- a/src/pymmcore_widgets/useq_widgets/_positions.py
+++ b/src/pymmcore_widgets/useq_widgets/_positions.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import json
+import warnings
 from dataclasses import dataclass
 from pathlib import Path
 from typing import TYPE_CHECKING, cast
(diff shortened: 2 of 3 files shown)


tlambert03:
tests are failing, but i think it might be the assertion in the test itself that is no longer correct

PR #382, opened 2024-11-06
feat: indicate pre-init properties in device property browser

### LINKED ISSUE #375: Feature: indicate pre-init properties in device property browser
it's possible to filter by pre-init props in the device property browser, but it's not otherwise apparent _which_ of the properties are pre-init (without toggling).  These could have a small 🅟 icon or something next to them

fdrgsp:
tests failing because of https://github.com/hgrecco/pint/issues/2074.

update for #375 .

@tlambert03 instead of using a specific icon as you mentioned in #375 , what about changing the background color as we do for the read only props?

Screenshot 2024-11-06 at 8 34 57 AM

Screenshot 2024-11-06 at 8 35 17 AM

closes #375

### DIFF
--- a/src/pymmcore_widgets/device_properties/_device_property_table.py
+++ b/src/pymmcore_widgets/device_properties/_device_property_table.py
@@ -5,7 +5,7 @@ from typing import TYPE_CHECKING, cast
 
 from pymmcore_plus import CMMCorePlus, DeviceProperty, DeviceType
 from qtpy.QtCore import Qt
-from qtpy.QtGui import QColor
+from qtpy.QtGui import QPalette
 from qtpy.QtWidgets import QAbstractScrollArea, QTableWidget, QTableWidgetItem, QWidget
 from superqt.fonticon import icon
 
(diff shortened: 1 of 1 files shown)


tlambert03:
that works for me.  did you happen to check how it looks on a dark theme?

tlambert03:
oh well :) it's fine.  we've got enough dark theme issues anyway.  if we ever coordinate theming more generally we can address it then

tlambert03:
however, if you're inclined, you could look into using a https://doc.qt.io/qt-6/qpalette.html#ColorRole-enum (which would pull from the palette) instead of hard-coding a color

src/pymmcore_widgets/device_properties/_device_property_table.py
```diff
@@ -123,8 +123,16 @@ def _rebuild_table(self) -> None:
...
+
+        # Retrieve the palette
+        palette = self.palette()  # This pulls the current theme colors
+
+        read_only_color = palette.color(QPalette.ColorRole.Dark)
+        pre_init_color = palette.color(QPalette.ColorRole.Dark).darker(130)
+
         for i, prop in enumerate(props):
-            item = QTableWidgetItem(f"{prop.device}-{prop.name}")
+            extra = " (pre-init)" if prop.isPreInit() else ""
```
tlambert03, line 134: i'm afraid of the manual styling not working.  I think I would pretty much remove all the styling from this PR and just keep this line, but use:

```suggestion
            extra = " 🅿" if prop.isPreInit() else ""
```

fdrgsp: > i'm afraid of the manual styling not working in some (as of yet unseen) circumstance. 

Screenshot 2024-11-06 at 9 48 50 PM

Issue #336, opened 2024-07-16
Make it clear that GroupPresetTableWidget slider values can be edited


gselzer:
* pymmcore-widgets version: 0.7.2
* Python version: 3.12.4
* Operating System: Windows 11

### Description

As I alluded to in [this comment](https://github.com/pymmcore-plus/pymmcore-widgets/pull/334#discussion_r1679930465), my colleague did not notice that the `QDoubleSpinBox` in the `GroupPresetTableWidget` could be edited. It'd be great to provide some visual clue (e.g. spin buttons) to indicate that it's editable, and not just a label.

![image](https://github.com/user-attachments/assets/9d45e548-c0cb-4d3b-b28c-6d5d381050b7)

tlambert03:
that's a widget from superqt.  
I'm a little bit conflicted on that one actually.  Those are ugly and take up valuable space; and once you notice they're editable, you know it forever.  
I can show you how to enable them via private methods (and we might want to add a public feature for that at superqt); I can see that someone might want it as an option;  but I'm not immediately sure I want to make it the default appearance for everyone.

gselzer:
> once you notice they're editable, you know it forever.

That's a fair point. There are a couple other things I can think of that could be helpful:
* Add a tooltip to the SpinBox that somehow implies editability
* Edit the border style of the SpinBox - that could be done within my UI, no change needed here, right?

marktsuchida:
I was thinking the same things. A tooltip on the combined slider-box widget could indicate which property it controls (as with the regular config groups), plus mention "click the number to enter a value".
Also, we should probably watch more people use it before concluding that we absolutely need to emphasize editability here.

tlambert03:
yeah, i think this is something that the app-builder should do with style sheets.  superqt does currently setStyleSheet on the slider label, but it shouldn't... and after https://github.com/pyapp-kit/superqt/pull/254, you could do this yourself with a stylesheet:


```python
from qtpy.QtWidgets import QApplication
from superqt import QLabeledSlider

app = QApplication([])
app.setStyleSheet(
    "SliderLabel { border: 2px solid black; background-color: yellow; }"
)
sld = QLabeledSlider()
sld.show()
app.exec()
```

<img width="430" alt="Screenshot 2024-07-17 at 9 53 59 AM" src="https://github.com/user-attachments/assets/cd17e48a-576a-44df-8152-c7f15007a9c5">

PR #383, opened 2024-11-07
feat: Absolute stage position controls

### LINKED ISSUE #332: Stage Control - consider absolute position field

gselzer:
This PR is designed to resolve #332.

### DIFF
--- a/src/pymmcore_widgets/control/_stage_widget.py
+++ b/src/pymmcore_widgets/control/_stage_widget.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from itertools import product
-from typing import cast
+from typing import TYPE_CHECKING, cast
 
 from fonticon_mdi6 import MDI6
 from pymmcore_plus import CMMCorePlus, DeviceType, Keyword
@@ -10,6 +10,7 @@ from qtpy.QtWidgets import (
     QCheckBox,
     QDoubleSpinBox,
     QGridLayout,
+    QHBoxLayout,
     QLabel,
     QPushButton,
     QRadioButton,
@@ -19,6 +20,9 @@ from qtpy.QtWidgets import (
 )
 from superqt.fonticon import setTextIcon
 from superqt.utils import signals_blocked
+
+if TYPE_CHECKING:
+    from typing import Any
 
 CORE = Keyword.CoreDevice
 XY_STAGE = Keyword.CoreXYStage
@@ -69,6 +73,28 @@ class MoveStageButton(QPushButton):
         )
 
 
+class MoveStageSpinBox(QDoubleSpinBox):
+    """Common behavior for SpinBoxes that move stages."""
+
+    def __init__(
+        self,
+        label: str,
+        minimum: float = -99999,
+        maximum: float = 99999,
+        *args: Any,
+        **kwargs: Any,
+    ) -> None:
+        super().__init__(*args, **kwargs)
+        self.setToolTip(f"Set {label} in µm")
+        self.setSuffix(" µm")
+        self.setMinimum(minimum)
+        self.setMaximum(maximum)
+        self.setDecimals(1)
+        self.setAttribute(Qt.WidgetAttribute.WA_MacShowFocusRect, 0)
+        self.setButtonSymbols(QSpinBox.ButtonSymbols.NoButtons)
+        self.setAlignment(Qt.AlignmentFlag.AlignCenter)
+
+
 class HaltButton(QPushButton):
     def __init__(self, core: CMMCorePlus, parent: QWidget | None = None):
         super().__init__("STOP!", parent=parent)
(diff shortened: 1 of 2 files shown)


gselzer:
I am especially interested in the reviewers' opinions on a036481, which fixes/improves the testing of setting absolute position. Allowing error does not make sense to me, but it is also necessary for the tests to pass (although the positional error is much smaller than the bounds I wrote)...

tlambert03:
Thanks @gselzer, the code looks good to me.  My main concern, and the reason I backed down from adding this in https://github.com/pymmcore-plus/pymmcore-widgets/pull/334 is that this introduces a very real danger to damage hardware.  programmatic control of motors, particularly when the objective is focused on a sample, is one of the main causes of expensive damage.  I did like @marktsuchida's suggestion in https://github.com/pymmcore-plus/pymmcore-widgets/pull/334#issuecomment-2233371248 that this functionality might be hidden behind an additional button and dialog that would prevent anyone doing this by accident.  I also think it would be great if that dialog also had a big "HALT" button that would stop and existing stage movement.  (the diSPIM plugin has one of those in micro-manager and i love it and use it all the time)

gselzer:
> My main concern, and the reason I backed down from adding this in https://github.com/pymmcore-plus/pymmcore-widgets/pull/334 is that this introduces a very real danger to damage hardware. programmatic control of motors, particularly when the objective is focused on a sample, is one of the main causes of expensive damage.

This is a great point - happy to refactor this PR so that we can resolve your concerns.

> I did like @marktsuchida's suggestion in [#334 (comment)](https://github.com/pymmcore-plus/pymmcore-widgets/pull/334#issuecomment-2233371248) that this functionality might be hidden behind an additional button and dialog that would prevent anyone doing this by accident.

This is one option. A second, simpler option is just an opt-in toggle to enable these spinboxes - otherwise they could just perform like the label that you had before. I actually thought about adding that as a part of this PR, but opted for simplicity. Happy to delegate to your and @marktsuchida's intuition here.

> I also think it would be great if that dialog also had a big "HALT" button that would stop and existing stage movement.

There's already [code for that](https://github.com/pymmcore-plus/pymmcore-widgets/blob/7668eda444e8dfff5a6ef665165b541901fb6cd3/src/pymmcore_widgets/control/_stage_widget.py#L72) although we could maybe create a smaller button with an icon instead of text, and put it alongside these spinboxes.

tlambert03:
> A second, simpler option is just an opt-in toggle to enable these spinboxes - otherwise they could just perform like the label that you had before.

I'm good with that option 👍... I wonder about spacing though, the X/Y case is already pretty wide horizontally, and has 4 checkboxes below. 

Screenshot 2024-11-08 at 3 39 23 PM

just an annoying aesthetic issue :) if you can come up with a nice look for it, let's go with that option

tlambert03:
hey @gselzer, just checking in here. let me know if you want any more ideas or just need time to implement.  I'd like to see a screenshot of how the buttons look when added

gselzer:
> hey @gselzer, just checking in here. let me know if you want any more ideas or just need time to implement. I'd like to see a screenshot of how the buttons look when added

Oh! Sorry, lost track of this - I'll take a look tomorrow once I have some time

gselzer:
@tlambert03 with 5034319 the UI looks like this:

image

Thoughts?

tlambert03:
I was expecting a button somewhere that toggles whether the XY positions are editable in the first place (with the default being not editable, the user should have to do some explicit action to enter a mode where the stage will move in response to number entry).

marktsuchida:
> * We'd then have to add another control - I feel a bit bad about all of the clutter I've already added 😅

That was one of the motivations for proposing that an additional dialog open. Alternatively, the text can be uneditable by default, but a small popup could open on right-clicking and selecting "Edit...". Perhaps.

fdrgsp:
>  the text can be uneditable by default, but a small popup could open on right-clicking and selecting "Edit...

I like this idea!

marktsuchida:
Of course this small popup would need to stay open (until something else is clicked?) if we want to allow people to edit it multiple times without re-clicking. Also, perhaps something like "Enter Coordinates..." would be a better name than "Edit...".

gselzer:
@fdrgsp I do like that implementation - are you able to push that to this branch?
Couple points:
* Should we tether the two spin boxes together on a 2-axis stage, such that both become enabled/disabled together?
* I like @marktsuchida's suggestion of  `Edit...` for the menu text, but then it should probably change to something else when enabled. Maybe `Enable Editing`/`Disable Editing`?

gselzer:
@fdrgsp thank you so much for contributing, in cf9942a I cleaned up your changes a bit, and added some tests. Notably, I:
* Tethered the spin boxes so enabling one enables both.
* Made the menu action toggle, so no need to change text.
* More minor changes.
Lemme know if you see any issues

fdrgsp:
@gselzer I like this solution but after having a quick chat with @tlambert03 I do realize that it might be a little hard for a user to know that there is the option to enable manual editing of the coordinates. 

What about having a menu on top of the widget? It might be easier to "discover"...


https://github.com/user-attachments/assets/1c4678cc-0149-4fc6-86d0-c9306de96d94

gselzer:
> @gselzer I like this solution but after having a quick chat with @tlambert03 I do realize that it might be a little hard for a user to know that there is the option to enable manual editing of the coordinates.

You know, this reminds me of the discussion that we had in #336 - specifically "once you notice they're editable, you know it forever."

I have a *very weak* preference for the current behavior as:
* It requires one fewer control.
* The menu is physically closer to the widgets it affects (but this is likely easily resolved).
* This behavior maybe *shouldn't* be discovered and used unless you're a poweruser who really knows what they're doing.
All of these reasons are pretty minor, so I'm content with being overruled here if you guys like it better this way.

PR #384, opened 2024-11-09
fix: fix af per pos bug on prop changed

fdrgsp:
I found a bug in the `CoreConnectedPositionTable` `_update_autofocus_enablement` method when core `propertyChanged` is called.

to reproduce:
- check the `CoreConnectedPositionTable.af_per_position` checkbox (the AF column will be visible).
- set the `Core Autofocus` to `""`.
- the `CoreConnectedPositionTable.af_per_position` is now disabled but the AF column is still visible.
 
I fixed this and modified the test so this bug is exposed without this fix.


tlambert03 (approved):
thanks!

PR #385, opened 2024-11-10
feat: use QRadioButtons to select z_plan

fdrgsp:
Use radio button in the z plan widget (as in the grid plan widget)

@tlambert03 what do you think?

---
with this PR:
Screenshot 2024-11-09 at 9 32 12 PM

before:
Screenshot 2024-11-09 at 9 33 46 PM

### DIFF
--- a/src/pymmcore_widgets/mda/_core_z.py
+++ b/src/pymmcore_widgets/mda/_core_z.py
@@ -51,7 +51,7 @@ class CoreConnectedZPlanWidget(ZPlanWidget):
 
     def setMode(
         self,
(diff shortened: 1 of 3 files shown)


tlambert03:
looks great thanks.  made some minor visual tweaks.  One issue remains though: for some reason, if you click on the *words* range around and above/below.  it works fine.  If you click on top/bottom, however, it *does* switch, but the radio button doesn't switch 🤔

fdrgsp:
> looks great thanks. made some minor visual tweaks. One issue remains though: for some reason, if you click on the _words_ range around and above/below. it works fine. If you click on top/bottom, however, it _does_ switch, but the radio button doesn't switch 🤔

I'll have a look!

PR #386, opened 2024-11-10
fix: disable autofocus axis widget if no autofocus device is loaded/selected

fdrgsp:
This PR updates the logic for when the autofocus devices (**Autofocus Axis widget** and **Autofocus Per Position checkbox**)should be disabled to then emit a correct `useq.MDASequence`. 

On top of disabling them when an absolute z plan is selected, this PR adds:

- If no Autofocus device is loaded they are disabled (thus responding to the core `systemConfigurationLoaded` signal).

- They are also disabled if the core `AutoFocus` property is set to `None` (thus responding to the core `propertyChnaged` signal).

Screenshot 2024-11-11 at 6 10 51 PM 1

### DIFF
--- a/src/pymmcore_widgets/useq_widgets/_mda_sequence.py
+++ b/src/pymmcore_widgets/useq_widgets/_mda_sequence.py
@@ -442,6 +442,8 @@ class MDASequenceWidget(QWidget):
             for pos in value.stage_positions:
                 if pos.sequence and pos.sequence.autofocus_plan:
                     axis.update(pos.sequence.autofocus_plan.axes)
+                    # store the current state of the autofocus per position checkbox
+                    self._use_af_per_pos = True
         self.af_axis.setValue(tuple(axis))
         axis_text = "".join(x for x in value.axis_order if x in self.tab_wdg.usedAxes())
         self.axis_order.setCurrentText(axis_text)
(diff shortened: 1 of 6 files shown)


src/pymmcore_widgets/useq_widgets/_mda_sequence.py
```diff
@@ -509,54 +506,66 @@ def _settings_extensions(self) -> str:
...
         self.stage_positions.af_per_position.setToolTip(tooltip2)
+        # hide the autofocus columns if autofocus per position is disabled
         if not state:
-            self.stage_positions.af_per_position.setChecked(state)
-        else:
-            # re-enable autofocus per position only if it was checked before
-            self.stage_positions.af_per_position.setChecked(self._use_af_per_pos)
+            # not simply calling self.stage_positions.af_per_position.setChecked(False)
+            # because we want to keep the previous state of the checkbox
+            self.stage_positions._on_af_per_position_toggled(False)
```
tlambert03, line 519: this line smells a little funny.  we're using a private method of the stage_positions tab that looks mostly like an internal callback (that I wouldn't expect this other class to be using).  And it looks like we're just doing it to get that column hidden?  

though it might make for more code, one idea to keep these concerns more cleanly separated would be to add an eventFilter to the `PositionTable.af_per_position` Checkbox that listens for `QEvent.EnabledChange` events and hides the af column accordingly

tlambert03 (commented):
mostly looks good.  couple minor comments, and a (very mild) concern about `self.stage_positions._on_af_per_position_toggled(False)`

src/pymmcore_widgets/useq_widgets/_mda_sequence.py
fdrgsp: @tlambert03 I moved the logic in the `eventFilter` in the `CoreConnectedPositionTable`.

tlambert03 (approved):
looks good!

PR #390, opened 2024-11-22
refactor: move useq_widgets to new top level module

tlambert03:
this makes it even clearer that useq_widgets doesn't require or import from pymmcore stuff at all, and preps an eventual split into an independent package


tlambert03:
docs will need to be fixed for this

PR #409, opened 2025-03-07
[wip] info widget with image histogram, min and max values

simonecoppola:
Still work in progress - the idea is to have something similar to the Inspect window in Micro-Manager studio.

It could be linked to an `ImagePreview` widget to control clims - that's the next step as well as making it a bit nicer by including the option to control the range displayed/bins etc.

It uses `pyqtgraph` instead of `vispy` - I think in principle same result can be achieved with `vispy`

image

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -70,6 +70,7 @@ image = [
     # fix OSError: dlopen(Quartz.framework/Quartz on vispy
     "pyopengl; platform_system == 'Darwin'",
 ]
+plot = ["pyqtgraph"]
 
 test = [
     "pymmcore-widgets[image]",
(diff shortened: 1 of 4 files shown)


simonecoppola:
Ah I see, thanks for pointing me towards that!
If I use ndv for the histogram is there still scope for a dedicated widget like this? I think it would be quite a useful addition and I am not particularly bothered about using pyqtgraph specifically, so I'd be happy to rewrite it without using it to avoid adding a dependency

simonecoppola:
Yes, absolutely! I'll close this pull request as I think it's clear this is better off being just a custom tool for my use at the moment

PR #415, opened 2025-04-03
fix: fix plate calibration

fdrgsp:
I found two bugs in the `PlateCalibrationWidget` of the `HCSWizard`:

- The `y` positions stored in each `HoverEllipse` item should be negative since we use a coordinate system where y is positive upwards and negative downwards. This PR inverts the sign.

Screenshot 2025-04-02 at 8 17 14 PM
 
---

- It also fixes the slow behavior that I found when adding position from the HCSWidget (the `_on_rows_inserted` was unnecessarily called every time a row was added)


tlambert03 (approved):
looks good to me. merge whenever

tlambert03:
the test fails are weird, and i saw them happen another time this week.  try just re-running

PR #407, opened 2025-03-05
feat: stage explorer part 1

fdrgsp:
We decided to split the main Stage Explorer PR (#400) in sub-PRs for an easier review.

This is the first one.

This one just adds a basic wrapper around a SceneCanvas, with three methods:

- add_image(img, transform), to add a new image to the canvas
- zoom_to_fit(), to reset the zoom to contain the full scene
- clear(), clear all images on the canvas

PR #400, opened 2025-02-11
[WIP] feat: Stage explorer

fdrgsp:
updated version of #305 

In this PR I started thinking about a widget that tracks the stage movements and is able to record the positions (and move to new positions with double-click). The widget is named `StageExplorer`. To see an example of usage run the `stage_explorer_widget.py` example.

https://github.com/user-attachments/assets/b7a9250d-d1b0-462d-b70e-23fe78733a3f

PR #420, opened 2025-04-21
feat: stage explorer part 2

fdrgsp:
second part of #400 after #407


tlambert03:
i think [83c9e72](https://github.com/pymmcore-plus/pymmcore-widgets/pull/420/commits/83c9e72e774bdd1f0f50aad76f2e77d222fc2cc0) fixes the calculation of scene bounds with non-standard transforms

tlambert03:
merged @fdrgsp!  Carry on adding additional features from #400, like current position indication, area selection (i.e. conversion of some selection on the canavs to some useq object)

PR #423, opened 2025-05-04
fix: Stage controller objects that batch repeated relative/absolute calls and emit moveFinished when done

### LINKED ISSUE #410: Stage Widget "Snap on Click" -> "Snap after Move"
The stage widget has a CheckBox with label `"Snap on Click"`, which differs from mmstudio's `"Snap after Move"`. Some of our folks prefer the latter, as a common goal is "move the stage some amount in some direction, and then snap an image", which aligns more with "Snap after Move". 

I was going to change the label, however thinking more about `StageWidget._move_stage_relative`, this isn't actually the behavior that is implemented; @marktsuchida told me that `CMMCore.setRelativePosition` does not block, meaning that if you snap on click, you may be snapping an image prior to the move, during the move, even getting motion blur. We should change the behavior to align more with the intentions.

Of course, this goal is easier said than done. `StageWidget._move_stage_relative` is (currently) called on the GUI thread, so if we want to wait to snap an image we'll have to find some way to wait, ideally without blocking the GUI in the meantime. mmstudio seems to do this by maintaining a [map of stages to `Runnable`s](https://github.com/micro-manager/micro-manager/blob/58733515616259d71f578dd898e10253a5d27671/mmstudio/src/main/java/org/micromanager/internal/navigation/XYNavigator.java#L35) - a solution taking inspiration from that architecture may be wise.

And this brings up other fun questions, mainly "what if the user goes crazy on the button?" I assume that right now many stages throw errors on successive calls that are caught [here](https://github.com/pymmcore-plus/pymmcore-widgets/blob/37171860471fdf77dcfef466d5a8bb999af5ca72/src/pymmcore_widgets/control/_stage_widget.py#L464), while others may block the calling thread until the move completes. Our solution should be consistent *across devices* about:
1. What happens when a move is initiated and then a second move is requested by another click *before the first finishes*. We could:
  *  Ignore these calls/debounce. This seems most logical to me - naively (and I am naive) if I'm spamming a button like this, I generally just want to keep the stage moving the entire time I'm spamming, and the actual number of times I pressed the button is not terribly important.
  * Queue these calls into a future move. This is what mmstudio's `XYNavigator` [does](https://github.com/micro-manager/micro-manager/blob/58733515616259d71f578dd898e10253a5d27671/mmstudio/src/main/java/org/micromanager/internal/navigation/XYNavigator.java#L218-L227), and I'm guessing that there may be a reason for this. This solution promotes overscrolling - it could be cancelled with the "STOP" button, but feels like feature misuse. Would be interested in the thoughts of @nicost here, leaning on years of expertise with `XYNavigator` 😄 
2. When/how many snaps actually occur. If we debounce the behavior would be straightforward, although large pans would become jagged movements. Queueing more likely lends itself to a single snap at the end of all of the queued moves (and this is what mmstudio [does](https://github.com/micro-manager/micro-manager/blob/58733515616259d71f578dd898e10253a5d27671/mmstudio/src/main/java/org/micromanager/internal/dialogs/StageControlFrame.java#L801), however this likely also promotes overpanning.

All of this is to say that this is an involved fix, and there are likely tradeoffs. cc @tlambert03 @fdrgsp

tlambert03:
This is an attempt to fix #410 ... it's still a WIP, but it introduces a `_ValueBatcher` object, a generic object that knows how to get/set and add values, and check device busy state.  It manages repeated calls to `add_relative` and/or `set_absolute`, and, maintains an internal awareness of whether we're "done" or not.

That all sounds rather abstract, and it is, but the goal is to have one object that we can *share* (across, say, the stage controller widget, and the stage explorer widget #400 ... that coordinates all of the attempts to control and move the stage, and, importantly, lets them all know when the job is done so they can, for example, snap an image.

The value batcher itself doesn't manage an event loop, and could be moved into pymmcore-plus.  It is wrapped here with a small QObject that does have the timer, polls the controller, and emits the signal when needed.

Here is a concrete example of what it enables (there's still one little glitch that can be seen in the movie... and I'm not yet sure if that's a bug in this code, or the demo stage device adapter).  Note that you can press the relative move button twice (before it's done) and it still makes a 2x movement.

https://github.com/user-attachments/assets/670b1d37-bde2-427b-87ca-3cfdb4e848ca

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,7 +5,7 @@ ci:
 
 repos:
   - repo: https://github.com/crate-ci/typos
-    rev: v1
+    rev: v1.32.0
     hooks:
       - id: typos
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -27,5 +27,5 @@ repos:
       - id: mypy
         files: "^src/"
         additional_dependencies:
-          - pymmcore-plus >=0.13.4
+          - pymmcore-plus >=0.14.0
           - useq-schema >=0.5.0
--- a/examples/stage_explorer_widget.py
+++ b/examples/stage_explorer_widget.py
@@ -9,15 +9,9 @@ app = QApplication([])
 mmc = CMMCorePlus.instance()
 mmc.loadSystemConfiguration()
 
-# add a small rotation
-# mmc.setPixelSizeAffine("Res10x", (0.9962, -0.0872, 0.0, 0.0872, 0.9962, 0.0))
-# mmc.setPixelSizeAffine("Res20x", (0.4981, -0.0436, 0.0, 0.0436, 0.4981, 0.0))
-# mmc.setPixelSizeAffine("Res40x", (0.24905, -0.0218, 0.0, 0.0218, 0.24905, 0.0))
-
 # set camera roi (rectangular helps confirm orientation)
 mmc.setROI(0, 0, 400, 500)
 
-mmc.setProperty("XY", "Velocity", 1)
 
 explorer = StageExplorer()
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,7 +49,7 @@ classifiers = [
 dynamic = ["version"]
 dependencies = [
     'fonticon-materialdesignicons6',
-    'pymmcore-plus[cli] >=0.13.7',
+    'pymmcore-plus[cli] >=0.14.0',
     'qtpy >=2.0',
     'superqt[quantity,cmap,iconify] >=0.7.1',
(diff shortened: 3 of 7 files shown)


tlambert03:
hey @fdrgsp, would you mind banging on this for a little bit and trying to break it?

tlambert03:
this now uses pymmcore-plus v0.14 with https://github.com/pymmcore-plus/pymmcore-plus/pull/462

tlambert03:
this is looking pretty good.  here's a real scope (it's a bit extra laggy, because this was also over remove desktop):


https://github.com/user-attachments/assets/cc56bc5d-8532-4a8e-91af-a5cd700abd58

PR #425, opened 2025-05-05
build: setup uv dependency groups and use for testing

tlambert03:
updates pyproject to use uv


tlambert03:
@fdrgsp, you will likely want to pull and merge any open branches after this goes in.  (but it should make everything easier locally!)

PR #428, opened 2025-05-07
[WIP]: feat: add rois to the stage explorer

fdrgsp:


src/pymmcore_widgets/control/_stage_explorer/_stage_explorer.py
```diff
@@ -610,6 +687,154 @@ def _t_half_width(self) -> np.ndarray:
...
+            # create the ROI rectangle for the first time
+            roi = self._create_roi(canvas_pos)
+            self._undo_stack.push(InsertRoiCommand(self, roi))
+
+    def _create_roi(self, canvas_pos: tuple[float, float]) -> ROIRectangle:
+        """Create a new ROI rectangle and connect its events."""
+        roi = ROIRectangle(self._stage_viewer.view.scene)
+        roi.visible = True
+        roi.set_selected(True)
+        roi.set_anchor(roi._canvas_to_world(canvas_pos))
```
tlambert03, line 724: i made some additional cleanups on ROI.  Next thing I see that we should do is here: we should pass the canvas position into the creation of the ROIRectangle, so that the very first place that the handles are positioned is correctly in world space.  (i.e. "please create an `ROIRectangle` *at* this position`), to avoid a flash of seeing the handles at the origin, and then quickly moving them when the mouse drags

tlambert03:
did a number of updates here, added an ROI manager, and a "scan selected ROI" button. movie added to original post

PR #431, opened 2025-05-13
Q roi manager

tlambert03:
this work stems from (and will probably supersede) the awesome work that @fdrgsp did in #428 

It extracts all of the roi-handling logic into a couple new objects, including a dataclass-based pure python ROI model, and a QAbstractListModel to manage a bunch of `ROI` objects, and make sure that everything is in sync, even if multiple "views" onto the model are used (such as a QListView to show a textual representation of all the rois in use).

It also gives a preview of the actual fields of view that would be scanned given a particular roi (shown as little rectangles filling the area).

It also implements a stub for a Polygon which can ultimately be integrated with the work @rlokkie is doing in https://github.com/pymmcore-plus/useq-schema/pull/224).

All together, it looks something like this:

https://github.com/user-attachments/assets/dd5c5b78-0eb4-4199-998f-6f56556d6a8e


tlambert03:
> Is the y-axis in the canvas correct?

I was just reading that document yesterday as well, and... possibly: no :) but let's address that more systematically in another issue.

Issue #372, opened 2024-10-14
TODO: Mouse wheel should not affect PropertyWidgets when they are embedded in a PropertyBrowser


tlambert03:
as you scroll through the list, you shouldn't be able to accidentally change the value

PR #432, opened 2025-05-13
fix: Disable preset combo scroll

gselzer:
Yesterday in the lab we accidentally ran into #372, which motivated me to fix the bug. I did see #307, however apparently that PR does not fix the problem.

This PR overrides the `wheelEvent` method to ignore `QWheelEvent`s on preset combo boxes, and adds a test ensuring as such.

### DIFF
--- a/src/pymmcore_widgets/control/_presets_widget.py
+++ b/src/pymmcore_widgets/control/_presets_widget.py
@@ -4,11 +4,20 @@ import warnings
 
 from pymmcore_plus import CMMCorePlus, DeviceType
 from qtpy.QtCore import Qt
-from qtpy.QtGui import QBrush
+from qtpy.QtGui import QBrush, QWheelEvent
 from qtpy.QtWidgets import QComboBox, QHBoxLayout, QWidget
 from superqt.utils import signals_blocked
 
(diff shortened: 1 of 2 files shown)


tlambert03:
oh awesome.  please try to fix sliders too! :) i swear to god i've tried to fix it so many times and failed.  i don't know why I struggled so much :joy:  anyway thanks for this.  will take a final look later today

tlambert03:
ok, this is probably better than nothing, but what I really wanted in #372 was for the mouse event to pass up to the scroll area.  with this approach, if you are scrolling and hit a combobox, then scrolling just stops (though, thankfully, no values are changed).  Also, it still leaves things like sliders changing, so doesn't quite close #372.  

I think we might have more global luck (and finer tuned control) by installing an event filter on the parent widget: https://doc.qt.io/qt-6/eventsandfilters.html#event-filters ... it *should* be able to intercept all events and decide whether or not to pass them down.  (though, i feel like I've struggled in the past to get it to work).  let me try something quickly

tlambert03:
k, i'm gonna merge this as an improvement, but i removed the closes flag, sinceI  consider that issue to still be unsolved while sliders are still sliding

PR #433, opened 2025-05-14
fix: Use an event filter to avoid scrolling

gselzer:
Following up on #432, @tlambert03 mentioned that what would be better than just ignoring scroll events would be to redirect them to the parent, since usually users want to scroll the table and not the contents. This PR uses an event filter to do just that.

Unfortunately, the sliders are still tricky, and I had to access the private `_slider` attribute of the `SliderProxy` to get it to work (as @tlambert03 foreshadowed in #432). This PR could be easily cleaned up and merged as is, but I'd prefer to spur discussion on how we make those pesky sliders behave.

Oh, and I need to fix the test I wrote for #432.

closes https://github.com/pymmcore-plus/pymmcore-widgets/issues/372

### DIFF
--- a/src/pymmcore_widgets/_util.py
+++ b/src/pymmcore_widgets/_util.py
@@ -7,15 +7,19 @@ from typing import TYPE_CHECKING, Any
 import useq
 from psygnal import SignalInstance
 from pymmcore_plus import CMMCorePlus
-from qtpy.QtCore import QMarginsF, QObject, Qt
-from qtpy.QtGui import QPainter, QPaintEvent, QPen, QResizeEvent
+from qtpy.QtCore import QEvent, QMarginsF, QObject, Qt
+from qtpy.QtGui import QPainter, QPaintEvent, QPen, QResizeEvent, QWheelEvent
 from qtpy.QtWidgets import (
+    QAbstractSlider,
+    QAbstractSpinBox,
     QComboBox,
     QDialog,
     QDialogButtonBox,
     QGraphicsScene,
     QGraphicsView,
     QLabel,
+    QScrollBar,
+    QTableWidget,
     QVBoxLayout,
     QWidget,
 )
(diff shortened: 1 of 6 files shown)


tlambert03:
we also really need this on the device property table (perhaps even more than the groups/presets editor).  that's what #372 was about.  Can you see if you can get this to apply when running `python examples/property_browser.py` and scrolling on a slider

tlambert03:
here's a self-contained QTableWidget that seems to do a good job preventing stuff bubbling up, without making assumptions about private stuff:

```python
class NoWheelTableWidget(QTableWidget):
    def __init__(self, *args, **kwargs) -> None:
        super().__init__(*args, **kwargs)
        self.viewport().installEventFilter(self)

    def setCellWidget(self, row: int, column: int, widget: QWidget) -> None:
        for child in widget.findChildren(
            (QAbstractSpinBox, QAbstractSlider, QComboBox)
        ):
            child.installEventFilter(self)
        widget.installEventFilter(self)
        super().setCellWidget(row, column, widget)

    def eventFilter(self, obj: QObject, event: QEvent) -> bool:
        if isinstance(event, QWheelEvent):
            # Scroll the vertical scrollbar manually, to avoid recursion error.
            if sb := self.verticalScrollBar():
                delta = event.angleDelta().y()
                sb.setValue(sb.value() - delta)
            # Consume the event so child widgets don't process it
            return True
        return super().eventFilter(obj, event)
```

it can be used as the base class for, e.g. `DevicePropertyTable` and anything else that needs it.  might need some more child checks too

tlambert03 (approved):
perfect!  This definitely closes #372 

is this read to go?  (and if so, in the future, can you please unmark your PRs as draft when you are happy to have them merged?  I read "draft" as "please do *not* merge this, I am still working on it and it's not yet ready for final review)

gselzer:
Oops, yes, ready to go!

PR #456, opened 2025-07-30
fix: don't check isinstance in stage controller

tlambert03:
related to https://github.com/pymmcore-plus/pymmcore-plus/issues/486 and  https://github.com/pymmcore-plus/pymmcore-remote/pull/11 

this is the one place we've found so far that actually checks isinstance instead of a structural check like hasattr, etc...
this one is trivial, and was an overly explicit catch of a possibly null weakref.  This change should allow pymmcore-widgets to be used with pymmcore-remote.  

Test added to prevent regressions.  

cc @gselzer.  let me know if you have a preferred alternative approach

### DIFF
--- a/src/pymmcore_widgets/control/_q_stage_controller.py
+++ b/src/pymmcore_widgets/control/_q_stage_controller.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import weakref
-from typing import ClassVar
+from typing import ClassVar, cast
 
 from pymmcore_plus import AbstractChangeAccumulator, CMMCorePlus, core
 from qtpy.QtCore import QObject, QTimerEvent, Signal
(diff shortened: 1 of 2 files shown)


gselzer:
Thanks @tlambert03! You beat me to filing a PR, as I was going to suggest [this changeset](https://github.com/pymmcore-plus/pymmcore-widgets/compare/main...gselzer:pymmcore-widgets:stage_accum_mmc)

tlambert03:
thanks.  that changeset holds a strong reference, which is one of the things we're trying to avoid with the weakref.  so I think i'll go with this `is not None` instead

gselzer:
> thanks. that changeset holds a strong reference, which is one of the things we're trying to avoid with the weakref. so I think i'll go with this `is not None` instead

Can you explain more about why the weakref is necessary?

tlambert03:
if we leave dangling references to cores (particularly in cached objects like this) then they never get cleaned up, and the c-level objects also won't get cleaned up (including open file handles for the logs, etc...)  It can lead both to memory leaks as well as tricky to debug issues.  cc @marktsuchida  who might have additional comments (we discovered some problems with this recently and fixed at least some of them ... see also https://github.com/pymmcore-plus/pymmcore-plus/pull/485 which fixes more on the core side)

PR #457, opened 2025-07-31
fix: Convert svg_path to posix

### LINKED ISSUE #440: Stage widgets icons on windows
In a windows environment, the stage icons result in a path joining error that is not present on MacOS.

There are a bunch of these that pop up and the icons are never shown.
qt.svg: Cannot open file 'C:/Users/opm/microscope-control/Usersopm
ppDataLocalpyconifysvg_cachemdi_chevron-triple-up_color_rgb(0, 180, 0)_1737398331.svg', because: The filename, directory name, or volume label syntax is incorrect.

The following silly patch fixes the issue but I'm sure there's a simple and real fix for long term.
```diff
--- a/src/pymmcore_widgets/control/_stage_widget.py
+++ b/src/pymmcore_widgets/control/_stage_widget.py
@@ -53,18 +53,24 @@
         self.setFlat(True)
         self.setFocusPolicy(Qt.FocusPolicy.NoFocus)
         self.setCursor(Qt.CursorShape.PointingHandCursor)
+
+        # Prepare stylesheet-safe paths *before* the f-string
+        # to avoid SyntaxError on platforms that might misinterpret backslashes.
+        path_default = str(svg_path(glyph, color="rgb(0, 180, 0)")).replace("\\", "/")
+        path_hover = str(svg_path(glyph, color="lime")).replace("\\", "/")
+        path_pressed = str(svg_path(glyph, color="green")).replace("\\", "/")
+
         self.setStyleSheet(
             f"""
             MoveStageButton {{
                 border: none;
                 background: transparent;
-                image: url({svg_path(glyph, color="rgb(0, 180, 0)")});
+                image: url({path_default});
                 font-size: 36px;
             }}
             MoveStageButton:hover:!pressed {{
-                image: url({svg_path(glyph, color="lime")});
+                image: url({path_hover});
             }}
             MoveStageButton:pressed {{
-                image: url({svg_path(glyph, color="green")});
+                image: url({path_pressed});
             }}
             """
         )
```

gselzer:
Have been running into #440 a fair bit lately. @bscott711 was on the right path with his fix, as it seems that qss does not like Windows paths (could not find an official reference for this with a quick search, but StackOverflow [suggests this](https://stackoverflow.com/a/26121846). The buttons now look normal on my machine.

Closes #440

### DIFF
--- a/src/pymmcore_widgets/control/_stage_widget.py
+++ b/src/pymmcore_widgets/control/_stage_widget.py
@@ -62,14 +62,14 @@ class MoveStageButton(QPushButton):
             MoveStageButton {{
                 border: none;
                 width: 38px;
-                image: url({svg_path(glyph, color="rgb(0, 180, 0)")});
+                image: url({svg_path(glyph, color="rgb(0, 180, 0)").as_posix()});
                 font-size: 38px;
             }}
             MoveStageButton:hover:!pressed {{
-                image: url({svg_path(glyph, color="lime")});
+                image: url({svg_path(glyph, color="lime").as_posix()});
             }}
             MoveStageButton:pressed {{
-                image: url({svg_path(glyph, color="green")});
+                image: url({svg_path(glyph, color="green").as_posix()});
             }}
             """
         )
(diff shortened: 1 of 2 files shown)


tlambert03 (approved):
yeah `as_posix()` is the right fix for qss.

thank you!

PR #463, opened 2025-08-18
fix: pass Group Preset mmcore to child widgets

gselzer:
~~TODO: Add tests~~ A test has been added, but I'm not in love with it. Testing in general is a bit tricky. I hesitate to directly patch `CMMCorePlus.instance` on a larger scale because we might want to test that reasonable things happen if we *don't* pass a `CMMCorePlus` instance to widgets elsewhere in the test suite. For that reason, I opted to patch `CMMCorePlus.instance` just in one test, and tried my best to extensibly include all public API within the same test. Of course there may always be future widgets that need to be blacklisted from this test (there are already some). Also open to leaving this test out, but it was useful as it found another bug I wasn't aware of.

### DIFF
--- a/tests/test_misc.py
+++ b/tests/test_misc.py
@@ -1,19 +1,7 @@
-from __future__ import annotations
-
-import inspect
 import re
-import warnings
 from pathlib import Path
-from typing import TYPE_CHECKING
-from unittest.mock import PropertyMock, patch
-
-from pymmcore_plus import CMMCorePlus
-from qtpy.QtWidgets import QWidget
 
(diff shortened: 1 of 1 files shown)


tlambert03:
Thanks Gabe.  Fixes are great, but I agree, the test isn't quite hitting it.  For one thing, it's not really testing the key problem (even if I remove the fixes in this PR, your test doesn't catch the problem that this PR is fixing).   We can work on a test in the future, but I don't want that to hold up the fixes here.  so I removed the test and will merge

PR #469, opened 2025-08-22
Catch missing core

tlambert03:
This follows up on #463, adding a test that catches a number of additional missed cases where the core wasn't passed through


tlambert03:
thanks @gselzer.  Yes it would be nice, but these are tricky cases for a reason, and I'm not yet ready to go down the path that your PR goes down.  The whole point of the new config model was to not need a constant core connection, but rather convert the core into pure model objects (so, we shouldn't need a constant core connection there just to get an icon... it's a leaky abstraction at the moment, but the fix is to grab the device type at model creation time, not to preserve the core).  we can get to it soon enough.  This PR is already a significant improvement, and I don't want to hold up the "good" for the "perfect".

Issue #472, opened 2025-09-17
Stage Widget: Snap on click when live view running throws error


gselzer:
Steps to reproduce:
1. Get a GUI. pymmcore-gui shows this nicely.
2. Load the demo configuration (`Devices->Load Demo Configuration`)
3. On the stage widget, toggle `Snap on Click`
4. Turn on Live view
5. Move the stage

You then get the following error:
```
RuntimeError: This operation can not be executed while sequence acquisition is running.

  File "C:\Users\gjselzer\code\pymmcore-plus\pymmcore-gui\.venv\Lib\site-packages\pymmcore_widgets\control\_q_stage_controller.py", line 89, in timerEvent
    cast("CMMCorePlus", core).snapImage()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "C:\Users\gjselzer\code\pymmcore-plus\pymmcore-gui\.venv\Lib\site-packages\pymmcore_plus\core\_mmcore_plus.py", line 1628, in snapImage
    self._do_snap_image()
    ~~~~~~~~~~~~~~~~~~~^^
  File "C:\Users\gjselzer\code\pymmcore-plus\pymmcore-gui\.venv\Lib\site-packages\pymmcore_plus\core\_mmcore_plus.py", line 1965, in _do_snap_image
    super().snapImage()
    ~~~~~~~~~~~~~~~~~^^

```

Should be an easy fix, I'd guess? Just don't snap if a (continuous) sequence acquisition is running?

PR #473, opened 2025-09-17
Don't snap if sequence is running

gselzer:
Solves #472


tlambert03 (approved):
i think this is the right choice here, thanks!

PR #483, opened 2025-11-29
fix: fix test failing because of PySide6

fdrgsp:
Not yet sure this is the best fix for the tests failing on PySide6 but I open the PR to check. 

I think is from v6.10 released in November, before that it was working.

after that we should `ci(dependabot)` PRs.


tlambert03:
i'm fed up with pre-commit.ci ... and will be ditching it, so don't worry about that failure here

PR #496, opened 2026-03-09
fix: update dialog flags and remove manual styling in ConfigWizard

fdrgsp:
- On Windows, without `ModernStyle` the wizard background is white; since the text is styled with white color, it is invisible. Fixed by setting `ModernStyle` on `ConfigWizard` (consistent with `HCSWizard`).
- Hardcoding colors is fragile across platforms/themes, so the active-step highlight now uses bold text only.
Screenshot 2026-03-09 at 3 19 03 PM
Screenshot 2026-03-09 at 3 19 52 PM

---
- On Windows, the dialog for setting an initial device property did not resize correctly to fit its content, and manual resizing was blocked by the `Qt.WindowType.MSWindowsFixedSizeDialogHint` flag. Fixed by overriding `sizeHint()` in `PropertyTable` to return an appropriate size.

Screenshot 2026-03-10 at 2 39 44 PM

### DIFF
--- a/src/pymmcore_widgets/hcwizard/config_wizard.py
+++ b/src/pymmcore_widgets/hcwizard/config_wizard.py
@@ -140,7 +140,7 @@ class ConfigWizard(QWizard):
         for i, label in enumerate(self.step_labels):
             font = label.font()
             if i == current_index:
-                font.setWeight(1000)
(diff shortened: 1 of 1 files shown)


tlambert03 (approved):
thanks!

PR #511, opened 2026-03-15
feat: enhance grid plan handling and x/y position management in MDASeq and PositionTable

tlambert03:
a bit more minimal alternative to #510 
closes #510

@fdrgsp I think I'm a bit more comfortable treating this job as a "purely visual" task.   I like that #510 disabled the X/Y position inputs when either a global or position-local grid plan was absolute.  But I disliked how it used the *data* model (i.e. the nullable float columns) to store that state.   This is a smaller alternative that still avoids ever generating an invalid MDASequence, and similarly shows it visually by enabling/disabling.  but no data mutations are done.

Screenshot 2026-03-15 at 1 22 15 PM

(and rather than seeing sentinel values, you just see the value that *would* be used if it were enabled again)

Screenshot 2026-03-15 at 1 22 44 PM

### DIFF
--- a/examples/config_groups_editor.py
+++ b/examples/config_groups_editor.py
@@ -1,0 +1,17 @@
+import sys
+
+from pymmcore_plus import CMMCorePlus
+from qtpy.QtWidgets import QApplication
+
+from pymmcore_widgets import ConfigGroupsEditor
+
+app = QApplication([])
+core = CMMCorePlus()
+core.loadSystemConfiguration()
+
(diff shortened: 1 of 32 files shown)


fdrgsp:
Great! Love it!

Just one comment...are we ok in maintaining the default value for x, y and z set to 0? Like if we set:

```py
seq = useq.MDASequence(stage_positions=[useq.Position(z=3)])
mda_wdg.setValue(seq)
```

the x and y values are set to 0?

PR #531, opened 2026-03-26
feat: add context menu for flipping the view in StageViewer

fdrgsp:


tests/test_stage_explorer.py
```diff
@@ -158,6 +158,47 @@ def test_stage_explorer_position_indicator(qtbot: QtBot) -> None:
...
+    qtbot.addWidget(viewer)
+
+    # initially no flip
+    assert viewer.view.camera.flip == (False, False, False)
+
+    # open context menu and toggle flip X
+    from qtpy.QtCore import QPoint
+    from qtpy.QtWidgets import QMenu
+
+    viewer._show_context_menu(QPoint(10, 10))
```
tlambert03: @fdrgsp, this PR is the one that's breaking all of our macos runners across all of pymmcore-plus org :joy:


it's this line right here... it's going to block forever because of the `exec()` call in _show_context menu.  (you have to mock exec to prevent that)

because the full org only has a limited number of macos runners... when this PR hangs indefinitely, it prevents all of our other repos from running tests

fdrgsp: 🤦‍♂️

PR #545, opened 2026-03-30
feat: use ConfigGroupsEditor in GroupPresetTableWidget.

tlambert03:
this is what @fdrgsp did in the christina branch

adds the new ConfigGroupsEditor as a button to the GroupPresetTableWidget... 
and adds an apply button that applies it to core ,and warns if you close without saving

image
image
image

closes #539 
closes #540


tlambert03:
that would be great!  wanna make that change to my branch and merge when ready?

PR #551, opened 2026-04-07
fix: synchronize loaded MDA channel group

gcharvin:
## Summary

When loading MDA channels, synchronize the selected channel group before setting the channel table row values.

This ensures the config/preset column is rebuilt with the correct preset choices before row data are applied. If the loaded sequence references a channel group that is not available in the current Micro-Manager configuration, the widget now emits a Python warning instead of silently showing mismatched presets.

## Background

This came from loading saved MDA settings on a TiEclipse configuration where the channel group/preset did not restore correctly at application startup/load time.

## Validation

- python -m ruff check src/pymmcore_widgets/useq_widgets/_channels.py tests/test_channel_widget.py tests/test_channel_group_widget.py
- QT_QPA_PLATFORM=offscreen python -m pytest tests/test_channel_widget.py tests/test_channel_group_widget.py -q: 4 passed
- python -m compileall -q src/pymmcore_widgets/useq_widgets/_channels.py

### DIFF
--- a/src/pymmcore_widgets/useq_widgets/_channels.py
+++ b/src/pymmcore_widgets/useq_widgets/_channels.py
@@ -51,6 +51,7 @@ class ChannelTable(DataTableWidget):
         # These will change in on_group_changed... so we store the current values.
         self._groups: Mapping[str, Sequence[str]] = {}
         self._config_column: ColumnInfo = self.CONFIG
+        self._loaded_group_warning: str | None = None
 
         # when a new row is inserted, call _on_rows_inserted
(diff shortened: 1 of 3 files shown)


gcharvin:
Yes, that works for me. I removed the tooltip/UI path and switched this to
a direct `warnings.warn()` when the loaded group is not available.

I agree that letting the composed GUI handle warnings globally is cleaner
than adding ad-hoc UI feedback inside the widget.
Best
Gilles

tlambert03 (approved):

PR #555, opened 2026-04-08
fix: "p" at the front of axis order when auto-inserting current stage position

### LINKED ISSUE #550: Question about axis-order without provided position
Hi all,

I’ve been experimenting with the `MDAWidget` and inspecting the sequences it generates.

I noticed that when no positions are explicitly provided, the current stage position is used, which makes perfect sense to me. However, I expected the position axis (`"p"`) to always appear at the front of the axis order, since conceptually you would first move to a position and then iterate over other axes (e.g., channels, time, etc.).

This seems to be the case when the time axis (`"t"`) is not present in the MDA sequence. In that situation, `"p"` is placed first (line 206). However, when `"t"` *is* present, `"p"` appears after `"t"` in the axis order (line 203-204).

Here are the relevant lines in the code:
<https://github.com/pymmcore-plus/pymmcore-widgets/blob/927f28dc2b38f1c60fe1f4c2d1564b5a762b004b/src/pymmcore_widgets/mda/_core_mda.py#L190-L207>

I’m curious about the reasoning behind this behavior. Am I misunderstanding how the `useq` model is intended to work, or is there a specific design consideration that I’m missing? Or is this a bug?

boydcpeters:
Ensure "p" axis is first when auto-inserting current position

When no stage positions are provided, the current position is added automatically and the "p" axis is placed at the front of axis_order.

Closes #550

### DIFF
--- a/src/pymmcore_widgets/mda/_core_mda.py
+++ b/src/pymmcore_widgets/mda/_core_mda.py
@@ -199,11 +199,8 @@ class MDAWidget(MDASequenceWidget):
             # not be in the event
             if "p" not in val.axis_order:
                 axis_order = list(val.axis_order)
-                # add the "p" axis at the beginning or after the "t" as the default
-                if "t" in axis_order:
-                    axis_order.insert(axis_order.index("t") + 1, "p")
-                else:
-                    axis_order.insert(0, "p")
+                # Always insert "p" at the front
+                axis_order.insert(0, "p")
                 replace["axis_order"] = tuple(axis_order)
 
         if replace:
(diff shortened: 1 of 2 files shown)


tlambert03:
we can't think of any critical reason against this.  I think the assumption was just to follow the more likely _multi-position_ pattern, where usually people want to visit each position per time point (as opposed to do the full t at p0 before moving to p1).  But of course, that distinction is moot anyway when there's a single position. So it really doesn't matter, and consistency seems better here.  thanks @boydcpeters

boydcpeters:
Ahh that makes sense, I can see how this came about when imaging multiple positions, where visiting each position for each timepoint is the more common pattern. Thanks for the clarification!

PR #558, opened 2026-05-26
Prevent PresetsWidget internal refreshes from applying configs

gcharvin:
## Summary
- Block `PresetsWidget` combo signals during the full internal preset-state refresh.
- Prevent adding/removing `<no match>` from triggering `setConfig()`.
- Add a regression test for the `<no match>` -> preset transition.

## Motivation
When `PresetsWidget` refreshes after core property changes, removing the currently selected `<no match>` item can emit `currentTextChanged`. Since `_on_combo_changed()` calls `mmc.setConfig()`, an internal UI refresh may accidentally apply a hardware preset.

This was observed during MDA channel switching, where transient non-matching device states could cause an unintended channel preset to be selected. The issue is generic Qt signal reentrancy, not microscope-specific.

There is no screenshot attached because this is not a visual rendering change: the problematic behavior is an internal Qt signal emitted during combo-box synchronization. The regression test captures the behavior directly by asserting that the UI refresh does not emit `configSet`.

## Fix
Wrap the entire `_update_if_props_not_match_preset()` combo synchronization in `signals_blocked(self._combo)`, including `removeItem`, `addItem`, and `setCurrentText`.

This matches the existing pattern used by `ChannelGroupWidget`.

## Tests
- `ruff check src/pymmcore_widgets/control/_presets_widget.py tests/test_presets_widget.py`
- `ruff format --check src/pymmcore_widgets/control/_presets_widget.py tests/test_presets_widget.py`
- `pytest tests/test_presets_widget.py tests/test_channel_widget.py tests/test_group_preset_widget.py -q`

### DIFF
--- a/src/pymmcore_widgets/control/_presets_widget.py
+++ b/src/pymmcore_widgets/control/_presets_widget.py
@@ -106,26 +106,29 @@ class PresetsWidget(QWidget):
     def _update_if_props_not_match_preset(self) -> None:
         if not self._mmc.getAvailableConfigs(self._group):
             return
-        for preset in self._presets:
-            _set_combo = True
-            for dev, prop, value in self._mmc.getConfigData(self._group, preset):
-                cache_value = self._mmc.getPropertyFromCache(dev, prop)
-                if cache_value != value:
-                    _set_combo = False
-                    break
-            if _set_combo:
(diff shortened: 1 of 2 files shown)


tlambert03 (approved):
looks good to me.  thank you!

